Issue #42, opened 2014-11-06
Any way to declare a default source code to avoid long  require(`../../../`)


rstormsf:
Any way to declare a default source code to avoid long  require(`../../../`) ?
Like browserify options
`paths: path.resolve('./src')`

PR #44, opened 2014-11-06
add baseSrcPath option to set the path for node to find source code module 

rstormsf:
#42


rstormsf:
Nope. Can't remember what was that.

PR #64, opened 2014-12-22
Add coverify transform after command-line transforms.

chromakode:
If coverify runs before a transform such as [reactify](https://github.com/andreypopp/reactify), it interferes with the JSX transform resulting in failure to run the tests. Moving the converify transform later in the process fixes running test files written in JSX (via `mochify --cover --transform reactify`).

Unfortunately, this does not change the transform ordering for browserify transforms specified in `package.json`. Combining reactify (via `package.json`) and `--cover` will still fail.

### DIFF
--- a/lib/mochify.js
+++ b/lib/mochify.js
@@ -117,10 +117,6 @@ module.exports = function (_, opts) {
   } else {
     b.plugin(phantom, opts);
   }
-  if (opts.cover) {
-    b.transform(coverify);
-    b.plugin(cover);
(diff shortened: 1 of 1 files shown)


mantoni:
Thanks for the patch and sorry for the late response. Was AFK on holidays.

mantoni:
Released with `2.1.1`.

chromakode:
No worries! Thanks for merging it. :)

PR #67, opened 2015-01-07
Added support for browserify options via subargs

wired8:
Simple extension to support browserify options via subargs.

### DIFF
--- a/lib/help.txt
+++ b/lib/help.txt
@@ -42,6 +42,8 @@ Options:
        --transform   Specifies a Browserify transform to add. Can be specified
                      multiple times. Options can be passed with subargs.
 
+      --browserify   Specify browserify options as subargs e.g. [--ignoreMissing true --debug true]
+
           --plugin   Specifies a Browserify plugin to add. Can be specified
                      multiple times. Options can be passed with subargs.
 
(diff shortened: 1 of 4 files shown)


lib/help.txt
```diff
@@ -42,6 +42,8 @@ Options:
        --transform   Specifies a Browserify transform to add. Can be specified
                      multiple times. Options can be passed with subargs.
 
+      --browserify   Specify browserify options as subargs e.g. [--ignoreMissing true --debug true]
```
mantoni: `--debug true` is always set by default, so this is missleading.

test/node-test.js
```diff
@@ -56,7 +56,7 @@ describe('node', function () {
     run('passes', ['--node', '--cover'], function (code, stdout) {
       var lines = stdout.trim().split(/\n+/);
       assert.equal(lines[0], '# node:');
-      assert.equal(lines[2], '  \u001b[90m.\u001b[0m');
+      assert.equal(lines[2], '  .');
```
mantoni: Interesting. Why did this ever work?

mantoni:
Thanks for the pull request. I'm happy to merge this feature.
I've added a comment about the help documentation. Can I ask you to also put the new option into the README.md?

mantoni:
Oh, and a test case would be fantastic!

mantoni:
Sorry for all the back and forth, but I removed the feature again after looking into writing some tests for it. The problem with this is that the given subargs are not interpreted by Browserify. They have to match the internal config property names. This is not obvious and might even requires reading the Browserify source.

I'd rather add additional options according to the Browserify CLI and map them to the corresponding properties. You are welcome to send a pull request if there is anything in particular that you are interested in. You earned contributor credits anyway ;-)

Thanks again.

PR #83, opened 2015-04-23
Adding browserifyOptions to arguments

gregorskii:
This will allow any browserify option to be set to the b instance
created by Mochify.

https://github.com/mantoni/mochify.js/issues/76

PR #88, opened 2015-05-27
passthru glob options so I can ignore files

jameslnewell:
I've structured my app specific code using [local npm modules](http://www.devworkflows.com/posts/using-local-npm-modules-in-npm-v2-0/). 

```
/
  index.js
  package.json
  test/
    index.js
  lib/
    my-local-module/
      index.js
      package.json
      test/
        index.js
  node_modules/ #I want to ignore all test files in this directory
    some-remote-module/
      index.js
      package.json
      test/
        index.js
```

I want to run all the test files in the local modules but not the test files in their dependencies. This PR passes through [glob options](https://www.npmjs.com/package/glob#options) to the `glob` module allowing me to ignore the test files I don't want to run.

```js
mochify('./assets/**/test/*.js', {glob: {ignore: './assets/**/node_modules/**/test/*.js'}});
```

---

In #83 you asked how a similar set of options could be specified from the command line. I'd suggest:

```bash
mochify "./assets/**/test/*.js" --glob.ignore "./assets/**/node_modules/**/test/*.js"
```

### DIFF
--- a/lib/mochify.js
+++ b/lib/mochify.js
@@ -66,7 +66,7 @@ module.exports = function (_, opts) {
     if (arg.indexOf('*') === -1) {
       entries.push(arg);
     } else {
-      Array.prototype.push.apply(entries, glob.sync(arg));
+      Array.prototype.push.apply(entries, glob.sync(arg, opts.glob || {}));
     }
   });
   if (!entries.length) {
--- a/test/api-test.js
+++ b/test/api-test.js
@@ -90,4 +90,20 @@ describe('api', function () {
     });
   });
 
+  it('should only run 1 test when glob options are passed', function (done) {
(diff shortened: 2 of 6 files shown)


mantoni:
Nice :) Fancy adding a basic test so that we make sure to not break the glob handling in the future?

jameslnewell:
Added a test that expects only 1 test to run instead of the normal 2 when the ignore option is passed to the glob module - is that suffice?

mantoni:
Yes, that is fine. Thanks. If you could now squash the commits, I'll merge.

mantoni:
Released in `v2.9.0`. Thanks!

jameslnewell:
ty!

PR #95, opened 2015-06-26
Only write to input if input stream is defined

scottcorgan:
If there are any syntax errors of any build steps that fail before it makes it to Phantomjs, Mochify stdouts a bunch of unnecessary stack trace data related to the input stream missing. This makes it tedious to track down that it was an error before my tests actually ran and not an error in my tests.

### DIFF
--- a/lib/phantom.js
+++ b/lib/phantom.js
@@ -46,14 +46,18 @@ module.exports = function (b, opts) {
 
     wrap.push(through(function (chunk, enc, next) {
       /*jslint unparam: true*/
-      input.write(chunk);
+      if (input) {
+        input.write(chunk);
+      }
       next();
     }, function (next) {
       if (opts.reporter !== 'xunit') {
         opts.output.write('# phantomjs:\n');
       }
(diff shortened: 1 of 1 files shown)


mantoni:
Cool, thanks! If you could add a small test, I'd merge and release a patch. It should be relatively easy to do. Let me know if I can help.

scottcorgan:
ya, not sure how to test this output in Mochify. Can you point me in the right direction?

mantoni:
You can add [a fixture](https://github.com/mantoni/mochify.js/tree/master/test/fixture) that causes an error to be emitted by browserify an then [run it like the other phantomjs tests](https://github.com/mantoni/mochify.js/blob/master/test/phantom-test.js#L115) to assert the output.

mantoni:
Sorry for leaving this open for too long.

mantoni:
So this PR completely fell under the table. I'm very sorry for that. This finally went out with version `2.13.1`. Thank you again.

PR #98, opened 2015-07-17
add missing output to opts.output

amchang:
The phantom output needs to print out somewhere.

I was scratching my head wondering why no output from my test logs was printing, there is that option already to set the output location, so just print it out now.

### DIFF
--- a/lib/mochify.js
+++ b/lib/mochify.js
@@ -162,7 +162,6 @@ module.exports = function (_, opts) {
   b.on('error', error);
   b.on('bundle', function (out) {
     out.on('error', error);
-    out.pipe(opts.output);
   });
 
   if (opts.watch) {
--- a/lib/node.js
+++ b/lib/node.js
@@ -57,5 +57,7 @@ module.exports = function (b, opts) {
 
   apply();
   b.on('reset', apply);
-
+  b.on('bundle', function (out) {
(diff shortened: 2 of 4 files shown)


mantoni:
Hmm, most PhantomJS related test cases now fail, some showing the output twice. Can you investigate?

amchang:
Yeah I'll take a look, so weird

mantoni:
Ok, now you're not piping the `min-webdriver` output anymore, right?

mantoni:
Oh, and it's still printing everything twice, but you changed the test cases to ignore that. *haha*

mantoni:
I suggest that you start over with a test case demonstrating the problem first, and then implement a solution that fixes the issue. Honestly, I don't see where the problem with the current implementation is.

For a cleaner history, I would also suggest this to be done in a new PR. What do you think?

amchang:
@mantoni I'm not sure what your looking for, are you looking for a test case of the error?

amchang:
Going to close for now ... I think I might have found the issue. Will reopen the PR if necessary

mantoni:
I was just surprised to hear that the main use case of this tool does not print anything at all for you. That is why I asked you to provide something that reproduces the issue.

Thanks for helping on Mochify! :+1:

PR #101, opened 2015-07-21
Add the option to ignore ssl errors

amchang:
Pass the new ignore-ssl-errors options to Phantomic.

Ignore PR#98, bottom line was there was an issue of how I was testing where there is no output if a test is running long with a high timeout, no progress indicator was running that a test is running, just a blank console. The fix would be to add an indicator something is running.

### DIFF
--- a/lib/args.js
+++ b/lib/args.js
@@ -16,6 +16,7 @@ var defaults = {
   debug     : false,
   wd        : false,
   recursive : false,
+  'ignore-ssl-errors': false,
   reporter  : 'dot',
   timeout   : '2000',
   port      : '0',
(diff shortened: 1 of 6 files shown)


mantoni:
Cool! Can you also bump phantomic to `^1.2.0` since this depends on the new feature.
Oh, and this sounds pedantic, but the help.txt does not fit in the print margin anymore. Most command line tools stay within the 80 character width to avoid weird line breaks. Thank you!

amchang:
Updated, hope the txt file looks alright now.

lib/args.js
```diff
@@ -69,6 +70,10 @@ function args(argv) {
     opts['web-security'] = opts['web-security'] === 'true';
   }
 
+  if (opts.hasOwnProperty('ignore-ssl-errors')) {
+    opts['ignore-ssl-errors'] = opts['ignore-ssl-errors'] === 'true';
```
mantoni: This seems unnecessary. When you invoke `mochify --ignore-ssl-errors`, then `opts['ignore-ssl-errors']` is already set to the boolean `true`. With this one is required to write `mochify --ignore-ssl-errors true` to make it work. See `--node` and `--wd`.

amchang:
@mantoni changed it to a boolean switch

mantoni:
Released in `2.11.0`. Thanks!

PR #103, opened 2015-07-28
pass colors through to mocaccino

boneskull:
does what it says on the tin

### DIFF
--- a/lib/mochify.js
+++ b/lib/mochify.js
@@ -102,7 +102,9 @@ module.exports = function (_, opts) {
     yields   : opts.yields,
     timeout  : opts.timeout,
     grep     : opts.grep,
-    invert   : opts.invert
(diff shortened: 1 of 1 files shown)


mantoni:
Cool! How about some basic tests and mentioning this in the docs?

boneskull:
@mantoni Is mentioning it in the docs necessary?  I figured this was a bug, since it's already mentioned in the docs, but doesn't actually work.

Where's an example of asserting arguments are passed to `mocaccino`?  l

mantoni:
Oh, sorry. I didn't pay enough attention. You're absolutely right, that's a bug. There are no specific tests for passing arguments to moccasino, "only" integration style test cases. Never mind.

mantoni:
Released in `2.11.1`.

PR #105, opened 2015-07-29
add postinstall script to use global phantomjs by default; closes #90

### LINKED ISSUE #90: phantomjs api option not working
I'm trying to set phantomjs directly with the api, but it doesn't seem to work with a local phantomjs installation.  
  
Perhaps I have the syntax wrong?  
```    
//local installation
var phantomjs = require('phantomjs');

mochify('./test/site/js/**/*.js', {
    transform: ['reactify'],
    reporter: 'spec',
    phantomjs: phantomjs
})
.bundle();
```  

However, It seems to work with the command line option `--phantomjs`, but I need to use the API.

boneskull:
- the script first checks to see if there is a local installation of package `phantomjs`
- if not, then it checks the PATH
- if it's not there, it goes ahead and installs it locally
- the `lib/phantom.js` module will default to using the global `phantomjs`, then fallback to local if it cannot be found.  `opts.phantomjs` overrides any of this.
- updated `README.md`
- add `which@^1.1.1` dependency

I wasn't sure how to test if the `phantomjs` executable was found.  since your tests in `test/phantom.js` are more of the integration variety, I figured it's covered if they pass.  Since Travis has it installed globally, we won't get coverage on a local install.  A solution may be to zap `process.env.PATH` and run the `phantom` tests a second time.

### DIFF
--- a/scripts/postinstall.js
+++ b/scripts/postinstall.js
@@ -1,0 +1,80 @@
+#!/usr/bin/env node
+
+'use strict';
+
+/**
+ * This script attempts to find:
+ *
+ * - a locally installed or linked PhantomJS
+ * - a PhantomJS in the PATH
+ *
+ * if neither of those are found, it will install PhantomJS locally.
+ */
+
+function log(msg) {
+  var args = Array.prototype.slice(arguments, 1);
+  args.unshift('=> ' + msg);
+  console.log.apply(console, args);
+}
+
+function installPhantom(callback) {
+  var exec = require('child_process').exec;
+
+  log('Attempting to install PhantomJS locally');
+  var install = exec('npm install phantomjs', function (err) {
+    if (err) {
+      return callback('Failed to install PhantomJS.  Do it manually');
+    }
+    log('Successfully installed PhantomJS.  To link it globally, ' +
+        'execute:\n\tcd node_modules/phantomjs && npm link');
+    callback();
+  });
+  install.stdout.pipe(process.stdout);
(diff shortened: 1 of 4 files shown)


boneskull:
You can see from the travis build that `phantomjs` was not installed:

```
25.36s$ npm install 
npm WARN optional dep failed, continuing fsevents@0.3.6
> spawn-sync@1.0.13 postinstall /home/travis/build/mantoni/mochify.js/node_modules/phantomic/node_modules/cross-spawn/node_modules/spawn-sync
> node postinstall
> mochify@2.11.0 postinstall /home/travis/build/mantoni/mochify.js
> node scripts/postinstall.js
=> Finding PhantomJS
=> PhantomJS not present locally; checking PATH
=> PhantomJS found in PATH
=> Mochify install complete!
```

scripts/postinstall.js
```diff
@@ -0,0 +1,80 @@
...
+
+  lstatPhantom(function (err) {
+    if (err) {
+      return log('Mochify installed with warning(s): ' + err);
+    }
+    log('Mochify install complete!');
+  });
+}
+
+if (require.main === module) {
```
mantoni: Just curious - what is the intention of this check?

mantoni:
Nice work. Thank you very much :+1:

mantoni:
Released in `2.12.0`.

scripts/postinstall.js
boneskull: mostly just to ensure the module cannot be `require`'d

Issue #93, opened 2015-06-03
Piping to out only breaks phantomJS


Swaagie:
Running mochify with basic options using phantomJS and browserify ~9.0.7, this change https://github.com/mantoni/mochify.js/commit/f221e5576ef551934ed60c1473948cfc0dd77b80 will just leave phantomJS dead in the water. Reverting the change will make it work again, perhaps pipe to both streams to make sure watch works?

Swaagie:
Excluded outdated browserify being the issue

mantoni:
Just released `v2.10.0` which now uses browserify 10. Can you check if this is still an issue? I can't reproduce.

If you still have this issue, please post the node and phantomjs versions you're using and how you're using Mochify (command line / API).

PR #111, opened 2015-08-12
[fix] revert the revert of f221e557

Swaagie:
For some reason the whole issue we had around https://github.com/mantoni/mochify.js/issues/93 is now solved by piping to `b.pipeline`. We don't have projects that use `--watch`, so you might want to double check on https://github.com/mantoni/mochify.js/commit/f221e5576ef551934ed60c1473948cfc0dd77b80

This works with phantomjs@2.0 and the latest browserify@11.0.1

As for the phantom test, that failed locally on 2.0.0, the error line seems to not be there.

### DIFF
--- a/lib/mochify.js
+++ b/lib/mochify.js
@@ -163,8 +163,8 @@ module.exports = function (_, opts) {
 
   b.on('error', error);
   b.on('bundle', function (out) {
-    out.on('error', error);
-    out.pipe(opts.output);
+    b.pipeline.on('error', error);
+    b.pipeline.pipe(opts.output);
   });
 
   if (opts.watch) {
--- a/lib/mochify.js
+++ b/lib/mochify.js
@@ -162,7 +162,7 @@ module.exports = function (_, opts) {
   }
 
   b.on('error', error);
-  b.on('bundle', function (out) {
+  b.on('bundle', function () {
     b.pipeline.on('error', error);
     b.pipeline.pipe(opts.output);
   });
--- a/test/phantom-test.js
+++ b/test/phantom-test.js
@@ -33,7 +33,6 @@ describe('phantom', function () {
       assert.equal(stdout.indexOf('# phantomjs:\n'
         + '1..1\n'
         + 'not ok 1 test fails\n'
-        + '  Error: Oh noes!\n'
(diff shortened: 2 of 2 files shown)


amchang:
This isn't quite the reversion in f221e55. How did piping from the internal pipeline change versus what's there currently?

Swaagie:
How is it not the reversion (aside from also listening to errors on that stream)? This will stream the results of tests to `stdout` normally again vs. having to console.log the output of the buffer after `bundle` has fully completed. 

From what I've seen the regression that makes our programmatic setup stop working is in the transition from 2.7 to 2.8, e.g https://github.com/mantoni/mochify.js/compare/v2.7.0...v2.8.0 and within that the only offending change is the from `b.pipeline` to `out`. For whatever reason 2/3 months back I was seeing exactly the opposite where that change actually made stuff stop working. There are quite the stream dependencies and some where updated, it could be some dependency regression. Anyways if we really want to get to the bottom of this, we need some unit test that triggers what we are seeing atm.

Swaagie:
Also it seems the test regression around removing the error line is around `1.9.8` vs `2.0.0`, see http://docs.travis-ci.com/user/build-environment-updates/2015-04-09/#Updates to we want to use the latest?

mantoni:
Thank you very much for all the information around the issue. I still don't understand why, but going back to `b.pipeline` still breaks `--watch` for me. I'm still on PhantomJS 1.9.8 though and will have to test if this behaves differently with 2.0.0.

I guess we will have to get to the root cause of the issue to understand how to fix it for good.

Swaagie:
@mantoni did you have a chance to test it against 2.0.0? Just got back from holiday myself, what route should we follow to get to root cause

mantoni:
Sorry for keeping this open for so long. There have been a few changes in the meantime that make it difficult to tell what the current status is.

I just confirmed that applying this change still breaks `--watch` for PhantomJS 1.9. I'm using Mochify also with PhantomJS 2 with no issues.

@Swaagie Please open a new issue if things are still not working for you and we might be able to find a different solution. Applying this path is not going to happen because it breaks `--watch` for my team.

PR #117, opened 2015-10-26
Pin Mocaccino to version 1.5 due to breaking change

swirlycheetah:
My unit tests are currently failing and the issue seems to point towards an issue with a recently released version of [mocaccino](https://github.com/mantoni/mocaccino.js), version 1.6.

This PR uses the tilde rather than the caret to pin the version to only take minor updates.

Hopefully once mocaccino is fixed this amendment can be reverted.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "browserify": "^11.0",
     "watchify": "^3.2",
     "coverify": "^1.0",
-    "mocaccino": "^1.5",
+    "mocaccino": "~1.5",
     "mocha": "^2.0",
     "phantomic": "^1.2.0",
     "consolify": "^2.0",


mantoni:
It would help if you could describe the actual issue and opened a ticket for mocaccino. If I'd merge, it would remove a feature that others may already depending on, causing other issues to be raised. Let's try and fix whatever is really broken.

mantoni:
Ah, you already did open https://github.com/mantoni/mocaccino.js/issues/13. Thanks!

swirlycheetah:
Sorry, I meant to link to the Issue raised in the initial description but forgot. Happy for this not to be merged if it's only an issue for me. I'm currently using a fork with this change in to enable my tests to run while we debug.

mantoni:
Ok, so this made it into `2.13.1` to fix the incompatibility. However, I already made `2.14.0` which explicitly depends on Mocaccino `^1.6` again and also explicitly bumps Mocha to `^2.3` to have a build that "officially" enables custom reporters. So in case you don't want to upgrade Mocha for some reason, make sure to stick with `2.13.x`.

PR #124, opened 2016-01-07
Handle browserify options for advanced use

### LINKED ISSUE #76: Use `paths` browserify option with mochify
Hey! 

I'm using the following in the browserify config: 
```js
browserify(
  entries: src + '/scripts/Main.js',
  dest: dest + '/js',
  outputName: 'Main.js',
  paths: ['./app/scripts/']
)
```

And I use the following in mochify config:
```js
mochify('./test/**/*.js', {
    reporter : 'spec',
    paths: ['./app/scripts/']
})
```

However, the `paths` option doesn't seem to be picked up. How can I pass browserify option to browserify? 

Thanks!

mateuszwrobel:
Resolves #76
This issue is connected with option "paths" which is only accessible via api so I didn't add cli option. It will allow to handle more complex scenarios. Please review and it would be nice if this part of code could soon be accessible via npm (really needed).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "mochify",
-  "version": "2.14.3",
+  "name": "mochify-browserify",
+  "version": "2.14.4",
   "description": "TDD with Browserify, Mocha, PhantomJS and WebDriver",
   "keywords": [
     "mocha",
@@ -48,7 +48,7 @@
   },
   "repository": {
     "type": "git",
-    "url": "https://github.com/mantoni/mochify.js.git"
+    "url": "https://github.com/mateuszwrobel/mochify.js.git"
   },
   "dependencies": {
     "browserify": "^11.0",
--- a/lib/mochify.js
+++ b/lib/mochify.js
@@ -88,17 +88,17 @@ module.exports = function (_, opts) {
     console.error(String(err) + '\n');
   }
 
-  var brOpts = {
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "mochify",
-  "version": "2.14.3",
+  "name": "mochify-browserify",
```
mantoni: I guess this changed by accident.

mantoni:
This PR is pretty much the same as https://github.com/mantoni/mochify.js/pull/83
Like I already mentioned there, I wouldn't want the command line options and the API to diverge. This does not change just because you "only" want it for the "path" option. Alternatively, we could support the "path" option explicitly.

Also, I'd like to have at least a basic test to make sure we don't break this feature in the future.

Lastly, the travis build is broken. Please investigate.

mateuszwrobel:
In browserify paths option isn't accessible via command line, so there is no point of adding this to CLI. We can add whole browserify-options via CLI, but it will look like horror.

mantoni:
Like I said, I'd rather support the `path` option explicitly then.

mateuszwrobel:
Ok, so I will do this soon with tests, thanks.

mantoni:
:+1:

mantoni:
Closing as the `--path` option was made available in version `2.17.0`.

PR #131, opened 2016-02-01
add --wd-file command line arg

kolodny:
relies on mantoni/min-webdriver#14


mantoni:
Good stuff. Now that min-wd 2.9.0 was released, I think this commit should also bump the min-wd dependency to `^2.9.0`.

kolodny:
done https://github.com/mantoni/mochify.js/pull/131/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R65

mantoni:
Released in `2.16.0`. :tada:

PR #132, opened 2016-02-15
Pass '--path' option to browserify.

jhytonen:
This is similar to pull requests https://github.com/mantoni/mochify.js/pull/124 and https://github.com/mantoni/mochify.js/pull/83, supporting only the browserify paths option.

### DIFF
--- a/lib/args.js
+++ b/lib/args.js
@@ -29,7 +29,7 @@ function args(argv) {
     string     : ['reporter', 'ui', 'phantomjs', 'consolify', 'timeout',
                   'port', 'yields', 'transform', 'plugin', 'grep', 'url',
                   'require', 'extension', 'web-security', 'bundle',
-                  'wd-file'],
+                  'wd-file', 'path'],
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mochify",
-  "version": "2.16.0",
+  "version": "2.16.1",
```
mantoni: Please revert this change. I'll bump the version in the release commit.

mantoni:
Thanks for the pull request. Nicely done. One question: Why is the option called `path` in Mochify and `paths` in Browserify? Shouldn't the names be aligned? The command line option could still be `--path`.

jhytonen:
@mantoni It is mainly because I followed convention. There is already a Browserify option called "extensions", which is "extension" in Mochify. It is right above the line that I added in mochify.js file.

mantoni:
Ok, cool. Please revert the package.json change and squash the commits. :+1:

jhytonen:
@mantoni Done.

mantoni:
Released in `2.17.0`. Thanks!

PR #133, opened 2016-02-15
fix require parameter add before other entry files

tamascsaba:
When i require other module only add after the tests.

### DIFF
--- a/lib/mochify.js
+++ b/lib/mochify.js
@@ -138,6 +138,14 @@ module.exports = function (_, opts) {
     b.plugin(phantom, opts);
   }
 
+  if (opts.require) {
+    [].concat(opts.require).forEach(function (r) {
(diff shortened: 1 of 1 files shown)


mantoni:
Hmm. [Browserify adds entries before requires](https://github.com/substack/node-browserify/blob/master/index.js#L88) and I would like to keep the same order.
However, they use the `require` function instead of `add`. Would that work for you as well?

tamascsaba:
I want to run scripts before my tests, only with ``add`` can you run it before.

mantoni:
Well, then you could just specify the scripts as arguments in the correct order:

```bash
$ mochify ./my-setup.js ./test/*.js
```

tamascsaba:
Nice tip :+1:

PR #136, opened 2016-04-26
Disable browserField for node tests

hakatashi:
It will be nice to have this. Codes from "browser" fields are [to complement the features not available in browser](https://github.com/defunctzombie/package-browser-field-spec#overview) and then not suitable for tests directly executed in Node environment.

Also, [browserify's `--node` option](https://github.com/substack/node-browserify#usage) disables this.

### DIFF
--- a/lib/mochify.js
+++ b/lib/mochify.js
@@ -98,6 +98,7 @@ module.exports = function (_, opts) {
   if (opts.node) {
     brOpts.builtins = false;
     brOpts.commondir = false;
+    brOpts.browserField = false;
     brOpts.detectGlobals = false;
     brOpts.insertGlobalVars = ['__dirname', '__filename'];
   }


mantoni:
Thank you! This makes sense, but there is a little confusing bit about what `--node` does in mochify. It's basically running browserify as if it was executed with `--bare` and pipes the result into a node process. Therefore it's not the same as the `--node` option for browserify, which is according to the docs an alias for `--no-browser-field`.

Unfortunately, I rely on the interpreation of the browser field, even when I run `--node` because I sometimes use node processes to perform load tests against a backend using the code that is normally executed in the browser.

Having said all that, I would support adding `--no-browser-field` instead of changing the current behavior. This would allow you to use it the way you want without breaking it for others.

Would you care to change this pull request accodingly and maybe add a test?

hakatashi:
I get it. No doubt it is evil to break the interpretation of browser field for others.

I don't mind changing (or maybe creating another) PR to support `--no-browser-field`. But adding tests about the browser field seems to be a bit complicated. Will it be something like this?

```js
// test/fixture/no-browser-field/test/no-browser-field.js
var dep = require('dep')
describe(...)

// test/fixture/no-browser-field/test/node_modules/dep/package.json
{"name": "dep", "main": "index.js", "browser": "browser.js", ...}
```

mantoni:
Yes, that sounds like a good approach. The tests are more integration tests, unfortunately.

hakatashi:
OK. I'll create another PR to implement this within a day or two. Thanks for your polite feedback!

mantoni:
Thanks! :+1:

Issue #137, opened 2016-04-27
Errors do not exit with code 1


jcrugzz:
When an error occurs within the test, I see a logged output of 

```sh
Error: Exit 1
```
But the process itself does not exit with code 1. This creates false positives in systems like jenkins that rely on the exit code to be accurate.

mantoni:
I use Mochify with Jenkins all the time and the exit code is 1 for failing tests.

Can you provide an example where this does not work as expected?

jcrugzz:
@mantoni i realized that in our own runner we weren't listening for the error event and doing `process.exit(1)`. The solution I came to locally before i discovered this was to add `process.exit(1)` to [here](https://github.com/mantoni/mochify.js/blob/master/lib/mochify.js#L89-L90). We can close this.

PR #138, opened 2016-04-27
[test] fix tests for new tap version that was pulled

jcrugzz:
Discovered this in the process of attempting to fix a "bug" in #137

### DIFF
--- a/test/phantom-test.js
+++ b/test/phantom-test.js
@@ -13,7 +13,7 @@ var run    = require('./fixture/run');
 
 
 describe('phantom', function () {
-  this.timeout(5000);
+  this.timeout(8000);
 
   it('passes', function (done) {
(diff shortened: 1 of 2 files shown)


mantoni:
Thanks. It's slightly worrying that the stack disappears. Any idea why?

jcrugzz:
@mantoni the stack didnt disappear, it just includes a longer stacktrace more thorough stacktrace. it didnt seem advantageous to test that particular string output as it causes false positive failures like i found here when a new version was installed. We dont need to be coupled that strictly to the string output, just acknowledge the passing or failing.

mantoni:
Sorry for keeping this open for so long. I was still stuck with PhantomJS 1.9 and therefore the test cases didn't fail for me and I couldn't really verify this properly. Thanks for looking into it!

PR #141, opened 2016-07-14
Add Test Coverage for --bundle and --consolify args 

### LINKED ISSUE #140: --bundle and --consolify don't play well with subfolders
The HTML runner generated using the `--consolify` flag does not resolve the relative path to the bundle generated by the `--bundle` flag; instead the absolute value is used.

### Steps to repro:
Run `mochify` over a test suite specifying both a `--bundle` and `--consolify` flag whose value points to a subfolder (eg: `tmp`).

```
mkdir tmp
mochify --recursive test/ --bundle tmp/tests.js --consolify tmp/runner.html
```

### Expected
* tmp/tests.js is created
* tmp/runner.html is created
* tmp/runner.html resolves the relative path to the javascript bundle and includes it via a script tag (ie: `<script src="./tests.js">`)

### Actual
* tmp/tests.js is created
* tmp/runner.html is created
* tmp/runner.html attempts to load the bundle using the exact value passed to the `--bundle` flag (ie: `<script src="tmp/tests.js>`)

jonnyreeves:
* Fail with a useful error if the user tries to use --bundle wihout --consolify
* Add test-coverage to check that both the consolify'd HTML and extracted bundle are written to disk

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -43,7 +43,6 @@ rules:
     no-return-assign: 2
     no-self-compare: 2
     no-spaced-func: 2
-    no-sync: 2
     no-throw-literal: 2
     no-trailing-spaces: 2
     no-undef-init: 2
--- a/lib/consolify.js
+++ b/lib/consolify.js
@@ -11,7 +11,6 @@ var fs        = require('fs');
 var through   = require('through2');
 var consolify = require('consolify');
 
-
 module.exports = function (b, opts) {
   consolify(b, {
     bundle: opts.bundle
--- a/lib/mochify.js
+++ b/lib/mochify.js
@@ -65,6 +65,11 @@ module.exports = function (_, opts) {
     }
   }
 
+  if (opts.bundle && !opts.consolify) {
+    console.log('--bundle must be used with --consolify option\n');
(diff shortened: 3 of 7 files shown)


mantoni:
Hey, thanks! One thing: The `bundle` option is, well, optional :) If it's not provided, the bundle is inlined into the html file. Can I ask you to ensure this continues to work?

jonnyreeves:
Ah sorry, I forgot to test that flow! Let me patch that up.

jonnyreeves:
Hey @mantoni; so in adding test coverage I realised that this was the wrong place to make this fix - I'll raise a PR against consolify to ensure the bundle is still written to the correct location, just the script src value is changed to be relative.

Cheers

.eslintrc
```diff
@@ -43,7 +43,6 @@ rules:
     no-return-assign: 2
     no-self-compare: 2
     no-spaced-func: 2
-    no-sync: 2
```
jonnyreeves: I could change this so it only affects tests if you would prefer?

mantoni:
Awesome! Thanks a lot 👍

PR #148, opened 2017-03-26
Adds support for exclude option to browserify

wesleytodd:
Not sure if there was a better way to do this, because there are plenty of unsupported options from browserify.  But this adds the one I needed.

### DIFF
--- a/lib/args.js
+++ b/lib/args.js
@@ -30,7 +30,7 @@ function args(argv) {
     string     : ['reporter', 'ui', 'phantomjs', 'consolify', 'timeout',
                   'port', 'yields', 'transform', 'plugin', 'grep', 'url',
                   'require', 'extension', 'web-security', 'bundle',
-                  'wd-file', 'path', 'external'],
+                  'wd-file', 'path', 'external', 'exclude'],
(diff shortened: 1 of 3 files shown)


mantoni:
Cool thanks. Mind adding a simple test? There are similar ones to copy and adopt.

wesleytodd:
Added.

mantoni:
This is a good test to add, but I was mainly thinking about an integration style test [like this very similar one](https://github.com/mantoni/mochify.js/blob/770be84c5e89f653ad822deb9747046f9634bf87/test/api-test.js#L95). 😃

mantoni:
I would still like to merge this if you could fix the conflict and add that unit test, @wesleytodd ?

wesleytodd:
Hey @mantoni, sorry I have been slammed lately and not able to update this.  I have at least two other things ahead of this on my priority list at the moment, so I don't want to promise anything.  Feel free to take my work without merging.  Or just amend my commit, which github then credits both of us.

mantoni:
Thanks for the update. I don't depend on this myself and I'm also quite busy atm. I'll just leave it open for now.

mantoni:
It's been more than a year. If anyone wants this feauture, please open a new pull request. Thanks!

PR #149, opened 2017-03-26
Fix behavior of --require flag

wesleytodd:
I was trying to use the require flag, and it was not working.  I believe this PR is the intended behavior, right?


mantoni:
That makes sense. However, the tests are failing now. Care to look into those?

mantoni:
Yes, you're right regarding the tests. Any help on improving them would be appreciated.

I was also thinking about making the tests more unit-test like. They're currently integration tests that launch a new process for each case which is very slow.

mantoni:
Released in `v3.1.1`. Thanks a lot! 👍

Issue #97, opened 2015-07-16
change viewport size in tests


balaclark:
Is there currently a way to get to the phantomjs page in order to change the `viewportSize` (http://phantomjs.org/api/webpage/property/viewport-size.html) before running tests? This would enable responsive code to be tested.

mantoni:
No, that's not yet implemented. I'd accept a PR on [phantomic](https://github.com/mantoni/phantomic) and a corresponding one here.

m90:
This feature is now available in `mochify@3.2.1`, you can use `--viewport-height` and `--viewport-width` to pass values.

PR #152, opened 2017-05-12
Add option to pass viewport sizing to PhantomJS

m90:
Would solve #97 

I made one mistake when doing the changes for that in [phantomic](https://github.com/mantoni/phantomic/pull/16) though, so we might want to fix this first? Will comment inline.


lib/phantom.js
```diff
@@ -48,12 +48,14 @@ module.exports = function (b, opts) {
...
-      phantomjs      : phantomPath,
-      'web-security' : opts['web-security'],
-      'ignore-ssl-errors': opts['ignore-ssl-errors']
+      debug              : opts.debug,
+      port               : opts.port,
+      brout              : true,
+      phantomjs          : phantomPath,
+      'web-security'     : opts['web-security'],
+      'ignore-ssl-errors': opts['ignore-ssl-errors'],
+      viewportWidth      : opts['viewport-width'],
```
m90: I did not respect the correct kebap-case everywhere when adding this to phantomic. We might want to fix that in there first, so we can get rid of that oddity in here.

m90: PR for that would be here: https://github.com/mantoni/phantomic/pull/17

mantoni: After changing this in phantomic, it needs to be updated here, right?

mantoni:
Released in `v3.2.0` 🎉 

I just noticed that we both forgot about the `help.txt` file which should list the new options also ... it's getting a bit longish now ... 🤔 

Another thing that came to mind, maybe for later: It would be cool to default the viewport values with the current actual console size – should we detect that we're running in a tty.

m90:
True that, I guess completeness trumps brevity so I will fix the help.txt real quick.

About the tty: cool idea, might be well worth opening an issue.

PR #160, opened 2017-11-07
Update to Mocha 4

fatso83:
Updates the Growl dependency that had a vulnerability
See
- https://snyk.io/vuln/npm:growl:20160721
- https://www.bithound.io/github/sinonjs/lolex/master/dependencies/npm/mochify#security-advisories

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,6 +248,8 @@ mochify().plugin(istanbul, {
 
 - Node 4.0+, 6.0+ (use `v2.x` for older versions)
 - PhantomJS 1.9, 2.0
+- v4.x
+    - Mocha ^4.0
 - v3.x
     - Mocha ^3.2
     - Browserify ^14.1
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,12 +2,25 @@
   "name": "mochify",
   "version": "3.3.0",
   "lockfileVersion": 1,
+  "requires": true,
   "dependencies": {
     "@studio/changes": {
       "version": "1.2.0",
       "resolved": "https://registry.npmjs.org/@studio/changes/-/changes-1.2.0.tgz",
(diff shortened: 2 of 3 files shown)


fatso83:
Not sure if it would have been preferable to move over Mocha to be a peer dependency, as it would probably be installed alongside Mocha in most cases, and that would avoid cases of bithound failing deep hierarchies. On the other hand: more cumbersome, probably, as "If Mocha should be a peer dependency, then why not x,y,z?".

mantoni:
Thank you! Upgrading mocha requires the corresponding change in mocaccino first:

https://github.com/mantoni/mocaccino.js/blob/master/package.json

Once that is release as a major it can be changes here for consistency.

It shouldn’t really be a dev dependency because you can use mochify as a replacement for mocha in a browser-only project. Also, that wouldn‘t help with upgrading major releases. We still need some level of control to make sure the API is compatible.

I’m not able to do this until later tonight. I‘ll give you push access to both repositories and npm.

fatso83:
@mantoni, I didn't mean they should be dev dependencies, but peer dependencies, so that they could all use the same version. You could still specify a version range in each project for the API compatibility, [like we do in sinon-test](https://github.com/sinonjs/sinon-test/blob/master/package.json#L47).

But hey, not important, I'll send a PR for `moccacino` as well :)

mantoni:
Like I explained, I'm using this project as the only test dependency without explicitly specifying mocha. Would a peer-dependency install the newest specified mocha version then?

Previous major releases came with some API changes anyway, so I guess this is the first time it would actually work to depend on a range. But let's not make this more complicated than it needs to be :smile: 

 Care to also update the compatibility list in the readme?

fatso83:
README and Mocaccino updated. Pushing new version to NPM once this is merged.

mantoni:
Awesome! 🎉 Thanks a lot.

PR #161, opened 2017-12-05
Use headless chromium

mantoni:
Outstanding work:

- [x] `executablePath` path option to launch a custom install
- [x] Make sure the `--debug` is usable and document how to
- [ ] Try to block navigation somehow. Maybe with event listeners that `preventDefault` or the like. There is a flag for PhantomJS, but this feature is not supported by Puppeteer.
- [x] Deprecate Node 4 as Puppeteer requires Node 6
- [x] Change descriptions and compatility notes in README.


mantoni:
I think this is good enough for now. People who depend on navigation being blocked should simply not update for now. This should go out because it's a huge improvement in compatibility, debugging and performance. Thoughts?

.travis.yml
```diff
@@ -1,5 +1,4 @@
 language: node_js
 node_js:
-  - "4"
   - "6"
-  - "7"
+  - "8"
```
boneskull: @mantoni out of curiosity, would you bump a major for this?

mantoni: Adding 8 would be a minor. Removing 4 is a major. Running builds on uneven node is basically just being curious about the upcoming release and has no say in versioning. I tend to only run builds on LTS versions these days.

mantoni:
@fatso83 No worries. I'll send a PR for Sinon in a few minutes 😄

mantoni:
Released in `v5.0.0`.

PR #163, opened 2017-12-19
Added a flag to allow chrome to run as root

### LINKED ISSUE #162: Option to pass additional arguments to chrome
I'm running mochify as the root user in docker. Chrome needs [special flags](https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#chrome-headless-fails-due-to-sandbox-issues) to allow it to run as root. It would be convenient if I could pass those in myself. 

I was picturing this taking one of two forms:

* `--chrome-as-root` (or a similar flag) - adds the two needed flags
* `--chrome-args [ insert flags here]` - allows adding of arbitrary flags

fearphage:
See https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#chrome-headless-fails-due-to-sandbox-issues

This passes  the mentioned flags above to chrome: `--no-sandbox --disable-setuid-sandbox`.

These are required to run Chrome as the root user. This fixes #162.

### DIFF
--- a/lib/args.js
+++ b/lib/args.js
@@ -34,7 +34,7 @@ function args(argv) {
       'viewport-height', 'outfile', 'https-server'],
     boolean: ['help', 'version', 'watch', 'cover', 'node', 'wd', 'debug',
       'invert', 'recursive', 'colors', 'ignore-ssl-errors', 'browser-field',
-      'commondir'],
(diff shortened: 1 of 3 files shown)


mantoni:
Released in `v5.1.0`. Thank you for the PR 🚀

Issue #165, opened 2018-01-21
Fail with meaningful error if server port is already taken


mantoni:
If the HTTPS server cannot be launched because the port is already taken does not yield any meaningful error message.

See https://github.com/sinonjs/sinon/issues/1643

m90:
Closed via #165.

Issue #164, opened 2018-01-21
Use random port if no port is specified


mantoni:
The `--https-server` option currently takes the port number where to open the HTTPS server. If no port number is given, Mochify should use a random free port. This is only possible if _no_ `--url` is given, or the URL must be parsed, the port number changed and the URL recreated with the random port.

The relevant code block is this one:

https://github.com/mantoni/mochify.js/blob/fd0c6b520caf0ade87289d66f111ef6a6f8f75fb/lib/chromium.js#L160-L191

See https://github.com/sinonjs/sinon/issues/1643

m90:
As per #168 passing no value to `--https-server` will now make mochify pick a random free port.

This still does not solve the situation where the user passes both `--https-server` **and** `--url`. Ideally the following would happen

> When specifying both, the --url and the --https-server options, it was / is the users responsibility to configure matching ports in both cases. This can stay as is for now. You'd basically have to specify the port that you're using in --url or it won't work.
>
> What would be awsome though, is if the automatic port feature could work with --url as well. I can imagine this working like this: We parse the --url for a port number and launch the server on the matching port. If the port is 0 (e.g. https://localhost:0/foo), we let the OS find a free port and change the URL to point to that port.

mantoni:
Support for random port selection if no port is given in `--https-server` was released in `v5.4.0`.

m90:
Closed via #169 and published to npm at `v.5.0`

PR #168, opened 2018-03-10
Improve port handling for https server

m90:
When no port is given for the --https-server arg, pick a random free one.
When a port is specified but already in use, fail with a meaningful error message.

This would solve #165 and #164.

I'm unsure about a few things, so I'll also leave comments in the diff.

### DIFF
--- a/lib/args.js
+++ b/lib/args.js
@@ -22,8 +22,7 @@ var defaults = {
   reporter: 'spec',
   timeout: '2000',
   yields: '0',
-  colors: null,
-  'https-server': null
+  colors: null
 };
 
 function args(argv) {
@@ -72,6 +71,10 @@ function args(argv) {
         opts[prop] = parseInt(opts[prop], 10);
       }
     });
+
+  if (opts.hasOwnProperty('https-server')) {
+    opts['https-server'] = parseInt(opts['https-server'], 10) || 0;
+  }
 
   if (opts.hasOwnProperty('web-security')) {
     opts['web-security'] = opts['web-security'] === 'true';
--- a/lib/chromium.js
+++ b/lib/chromium.js
@@ -158,10 +158,7 @@ module.exports = function (b, opts) {
   };
 
   var serverPort = opts['https-server'];
-  if (serverPort) {
-    if (!opts.url) {
-      opts.url = 'https://localhost:' + serverPort;
-    }
+  if (Number.isInteger(serverPort)) {
     server = https.createServer({
       key: fs.readFileSync(path.join(__dirname, '..', 'fixture', 'key.pem')),
       cert: fs.readFileSync(path.join(__dirname, '..', 'fixture', 'cert.pem')),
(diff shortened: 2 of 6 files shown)


lib/args.js
```diff
@@ -77,6 +76,10 @@ function args(argv) {
     opts['web-security'] = opts['web-security'] === 'true';
   }
 
+  if (opts.hasOwnProperty('https-server') && !opts['https-server']) {
+    opts['https-server'] = null;
```
m90: I think this is really not pretty. Is there another nicer way of saying "option yes, value no"?

lib/chromium.js
```diff
@@ -182,11 +216,21 @@ module.exports = function (b, opts) {
...
     });
-    server.listen(serverPort, function (err) {
+    allocatePort(serverPort, function (err, port) {
       if (err) {
         b.emit('error', err);
         finish();
+        return;
+      }
+      if (!opts.url) {
+        opts.url = 'https://localhost:' + port;
```
m90: Can adding this asynchronously create some race condition that I am not aware of?

lib/chromium.js
```diff
@@ -28,6 +29,42 @@ function broutHandler() {
   });
 }
 
+function checkPortAvailability(port, cb) {
```
mroderick: Instead of adding code that checks for available ports to this package, I would recommend using an existing package.

[`portscanner`](https://github.com/baalexander/node-portscanner) looks promising.

lib/args.js
mroderick: It's always going to be ugly when a variable is used for control flow and should also support a default value. Some refactoring of this could lead to splitting the two concerns, one variable to specify whether or not to start the server and one variable to specify which port to use.

mroderick: I'd try to not break the api for as long as possible.

There's nothing stopping us from creating more than one internal variable from just one CLI argument.

```js
var startServer = // was the https-server flag used?
var passedPort = // what was the value passed?
var tryOtherPorts = !passedPort;
var serverPort = tryOtherPorts ? defaultPort : passedPort;
```

lib/chromium.js
m90: Looks like this doesn't even need a dependency as `.listen()` will automatically assign a free port when you pass 0: https://nodejs.org/dist/latest-v8.x/docs/api/net.html#net_server_listen

m90:
@mantoni Did you have a chance to look at this yet? I'm still a bit unsure about what this should be doing when both `--url` and `--https-server` are being passed.

mantoni:
@m90 Sorry, I completely forgot. Thanks for pinging and for making this change. It's a very nice improvement.

I have additional thoughts regarding your question, but those don't have to be adressed in this PR.

When specifying both, the `--url` and the `--https-server` options, it was / is the users responsibility to configure matching ports in both cases. This can stay as is for now. You'd basically have to specify the port that you're using in `--url` or it won't work.

What would be awsome though, is if the automatic port feature could work with `--url` as well. I can imagine this working like this: We parse the `--url` for a port number and launch the server on the matching port. If the port is `0` (e.g. `https://localhost:0/foo`), we let the OS find a free port and change the URL to point to that port.

Like I said, I think it's something we could keep as an issue if we feel this would be a good idea. This PR is good to be merged, if you don't have anything to add.

m90:
I think this is a good idea. I will add a comment to #164 about how it is only partially done. I also think I should find the time to add it soon-ish.

Will release this as `5.4.0` now.

PR #169, opened 2018-03-21
Enable passing of url without additionally specifying port

m90:
When --url is given, the port to be used by the local https server can be derived in case the --https-server arg is empty.

This is part 2 of #164.

### DIFF
--- a/lib/args.js
+++ b/lib/args.js
@@ -76,6 +76,17 @@ function args(argv) {
   if (opts.hasOwnProperty('url')) {
     opts.url = url.parse(opts.url);
     delete opts.url.host;
+
+    if (opts.hasOwnProperty('https-server')) {
+      if (opts['https-server'] !== opts.url.port) {
+        console.log(
+          'The port values in --https-server and --url did not match,'
+          + ' got "' + opts['https-server'] + '" and "' + opts.url.port + '".\n'
+          + 'You should be able to use the --url option only.'
+        );
+        process.exit(1);
+      }
+    }
   }
 
   if (opts.hasOwnProperty('https-server')) {
--- a/lib/args.js
+++ b/lib/args.js
@@ -8,6 +8,7 @@
 'use strict';
 
 var subarg = require('subarg');
+var url = require('url');
 
 var defaults = {
   watch: false,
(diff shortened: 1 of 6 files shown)


lib/args.js
```diff
@@ -72,8 +73,15 @@ function args(argv) {
       }
     });
 
+  if (opts.hasOwnProperty('url')) {
+    opts.url = url.parse(opts.url);
+    delete opts.url.host;
```
m90: This is really terrible, but apparently it's what needs to be done when using the legacy `url`  API, see: https://nodejs.org/dist/latest-v8.x/docs/api/url.html#url_url_format_urlobject

The WHATWG URL API (which is much nicer) is supported in Node 7+, so we would need to add a polyfill or sth in here if we wanted to use it. Don't know if that's worth it.

test/args-test.js
```diff
@@ -93,9 +93,16 @@ describe('args', function () {
...
 
-    assert.equal(opts.url, 'localhost');
+    assert.equal(opts.url.href, 'https://localhost:8080/test.html');
+    assert.equal(opts['https-server'], 8080);
+
+    var withPort = args(
+      ['--url', 'https://localhost:8080/test.html', '--https-server', '4040']
+    );
+    assert.equal(withPort.url.href, 'https://localhost:8080/test.html');
+    assert.equal(withPort['https-server'], 4040);
```
mantoni: Hmm, is this what we want? Should this fail? We could also print a warning 🤔

m90: I thought about this too. Even with a warning this would probably have serious rabbit hole potential, so it's probably best to state that the given values conflict and fail.

mantoni:
Hey! Nice pull request. I only have the one inline comment.

Re the `url` thingy, I think it’s fine as is. It’s not so terrible actually. I don’t think it’s worth a polyfill.

test/args-test.js
mantoni: > Rabbit hole potential 

😄👍

m90:
Released this as `5.5.0` - I also showed some love for `help.txt` so that it's more in line with the README: https://github.com/mantoni/mochify.js/commit/8fa09119874808a3867fcaab14391ce989915460

mantoni:
Thanks a ton! Great stuff 👌

PR #170, opened 2018-05-04
Document async polling arg in readme and help text

m90:
Looks like this is not fully documented yet.


mantoni:
Yes thank you. It’s actually not appium specific. It has been built to work around polling issues with some selenium servers.

Anyway, I like the appium hint because that is currently a non-obvious adjustment one has to make. 👌

m90:
I tried using it today, and it's really neat. Nice addition!

Issue #171, opened 2018-05-10
Build is broken in Node 6 due to Chrome rejecting the SSL server's certificate


m90:
So I [fixed a typo in the README](https://github.com/mantoni/mochify.js/commit/ff62e65b3e956d0c1f0f8ad6196a1df2754ab39e) and noticed that the build is broken in Node 6 since [recently](https://github.com/mantoni/mochify.js/commit/2ac6322362712bf563fcbd527733a12213932065), most likely unrelated to the changeset (all changes since the last passing build in Node 6 have been README only).

It seems to reproducibly fail in Node 6 now due to errors like:

```
Uncaught AssertionError: 'This site does not have a valid SSL certificate! Without SSL, your site\'s and visitors\' data is vulnerable to theft and tampe == '# chromium:\n1..1\nlocation.href = https://localhost:8080/test.html\nok 1 url has H1 element\n# tests 1\n# pass 1\n# fail 0\n'
```

I don't really understand what makes Chrome reject the cert when using Node 6, when Node 8 passes, but this is what I found out until now:

- test will pass just fine for me locally running Node `v8.11.1`
- the [last passed build](https://travis-ci.org/mantoni/mochify.js/jobs/374972795) on Node 6 was apparently using the very same Node/npm versions as the [failing ones](https://travis-ci.org/mantoni/mochify.js/jobs/377430935), so it's likely a Chrome / Pupeteer change:

```
$ node --version
v6.14.2
$ npm --version
3.10.10
$ nvm --version
0.33.11
```

Interestingly passing `--ignore-ssl-errors` to the tests [does **not** silence the failures](https://travis-ci.org/mantoni/mochify.js/builds/377442545) (only 98% sure I am [doing this](https://github.com/mantoni/mochify.js/commit/e0d41e776581e8bbaf9166d63370759120841d7a) correctly though).

Unfortunately I don't have a Node 6 test suite of a real project up and running so I could test if this is something that requires action as it breaks the real world or if it's just a weird Travis glitch, but it's probably a good idea to keep an eye on it and try to drill down to what the real cause of this is.

m90:
I also [tried running this against more Node versions](https://travis-ci.org/mantoni/mochify.js/builds/377448022) and it looks like it's failing lt 8.

mantoni:
Thanks for looking into it. I tested locally with a fresh Chromium download and it worked fine on Node 6. I also ran it with `--allow-chrome-as-root` successfully. This seems to another Travis glitch.

[...]

PR #173, opened 2018-05-17
Filter out non-log messages when reading console

m90:
Puppeteer >= 1.4.0 introduced changes that will change the
way messages are passed to `console` events. To filter out
non-test related messages, filter all non-log messages. Also
bump puppeteer to ^1.4.0 and start testing in Node 10.

So this is fixing #171 - looks like we got bitten by: https://github.com/GoogleChrome/puppeteer/pull/2400 which changes what kind of messages will trigger a `console` event.

### DIFF
--- a/lib/chromium.js
+++ b/lib/chromium.js
@@ -98,6 +98,9 @@ module.exports = function (b, opts) {
 
       var consumer;
       page.on('console', function (msg) {
+        if (msg.type() !== 'log') {
+          return;
+        }
         var text = msg.text();
         if (text.indexOf('[EXIT ') === 0) {
           var code = text.substring(6, text.length - 1);
--- a/lib/chromium.js
+++ b/lib/chromium.js
@@ -98,10 +98,12 @@ module.exports = function (b, opts) {
 
       var consumer;
       page.on('console', function (msg) {
+        var text = msg.text();
         if (msg.type() !== 'log') {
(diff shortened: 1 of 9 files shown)


lib/chromium.js
```diff
@@ -98,6 +98,9 @@ module.exports = function (b, opts) {
 
       var consumer;
       page.on('console', function (msg) {
+        if (msg.type() !== 'log') {
```
m90: @mantoni is this safe to do? Would mochify itself ever use `error`  level?

m90:
I don't fully understand why my npm started locking all the `require`s in the lockfile to be quite honest. Is that something it should do?

mantoni:
I didn't quite get the new review workflow 🙈 I commented a few days back but they don't show until I set a review status and submit them... oh well.

Any non-log console output is not directly printed to stderr. The test cases where changed to distinguish between the two output streams and I've added some tests for `console.log`, `console.info`, `console.warn` and `console.error` to verify.

mantoni:
Shall we release this as a patch or a minor?

m90:
Thanks for adding the tests and `stderr` also seems like a smart thing to do.

Considering that it _also_ adds Node 10, I'd feel it could also be a minor. Should we release it like this?

mantoni:
Sure. Can I ask you to do it? I’m AFK now.

PR #174, opened 2018-06-04
Update min-wd to 2.11.0

SteffenL:
Allows mochify to take advantage of custom capabilities, which lets us specify browser options, set up BrowserStack integration, etc.

Would you like me to add some more helpful documentation to make people aware of the new features in min-wd? Do we need tests?

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ the test framework.
 - Run tests in real browsers
     - Supports [SauceLabs][] ([docs](#saucelabs-setup))
     - Supports [Appium][] ([docs](#appium-setup))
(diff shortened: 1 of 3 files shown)


mantoni:
Thanks! The update will be included in new installs automatically since we're using the `^` version range to pull in `min-wd`. It only makes sense to update explicitly and release a new version if we're advertising the new feature in the documentation or some additional integration would be needed.

Feel free to add an example to this PR and I'll happily merge.

mantoni:
Awesome! Thank you very much 👍

mantoni:
I've cut a new release to make this an explicit feature of Mochify. Released as `v5.8.0`.

Issue #177, opened 2018-08-08
Plugin mochify-istanbul fails to provide coverage if --url is specified


rafaelsortoxo:
Hi, good day.

I have noticed that the command line gets mixed up and the plugin mochify-istanbul fails to provide the correct coverage if we specify the --url for the file to use in the test cases.

I have created one repo that explains the problem [mochify-istanbul-plugin-issue](https://github.com/rafaelsortoxo/mochify-istanbul-plugin-issue)

If we execute the test script normally, mochify-istanbul correctly reports the coverage and notices that only 1 function is being evaluated.

```
~/SD_Card/mochify-issue$ npm run test
> mochify-issue@1.0.0 test /media/rafael/SD_Card/mochify-issue
> npx mochify --plugin [ mochify-istanbul --exclude **/*.spec.js --dir coverage --report text --report html ] *.spec.js

# chromium:


  index
    ✓ Should return the sum of 2 numbers


  1 passing (6ms)

----------------|----------|----------|----------|----------|----------------|
File            |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
----------------|----------|----------|----------|----------|----------------|
 mochify-issue/ |       80 |      100 |       50 |       80 |                |
  index.js      |       80 |      100 |       50 |       80 |              2 |
----------------|----------|----------|----------|----------|----------------|
All files       |       80 |      100 |       50 |       80 |                |
----------------|----------|----------|----------|----------|----------------|
```
If we then create a basic html file like the one included in the fixtures and specify the --url, even though the test continues to work, mochify-istanbul provides the report before waiting for the tests to actually run. There seems to be an out of order execution of options or something similar.

```
~/SD_Card/mochify-issue$ npm run test -- --url file:/tmp/index.html

> mochify-issue@1.0.0 test /media/rafael/SD_Card/mochify-issue
> npx mochify --plugin [ mochify-istanbul --exclude **/*.spec.js --dir coverage --report text --report html ] *.spec.js "--url" "file:/tmp/index.html"

----------|----------|----------|----------|----------|----------------|
File      |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
----------|----------|----------|----------|----------|----------------|
----------|----------|----------|----------|----------|----------------|
All files |      100 |      100 |      100 |      100 |                |
----------|----------|----------|----------|----------|----------------|

# chromium:


  index
    ✓ Should return the sum of 2 numbers


  1 passing (5ms)
[...]

PR #178, opened 2018-08-09
 Ensure browserify plugins are always mounted in the same order

m90:
This is an attempt at fixing #177.

Tests and all seem to pass and it also fixes the issue in the example repo provided in the issue, then again I am slightly worried if it is even a good idea to potentially add browserify plugins asynchronously or if this needs some bigger refactoring in order to do everything synchronously always.

### DIFF
--- a/lib/mochify.js
+++ b/lib/mochify.js
@@ -162,6 +162,10 @@ module.exports = function (_, opts) {
       next();
     } else if (opts.node) {
       b.plugin(node, opts);
+      // As opposed to when using the chromium / webdriver plugins
+      // `next` needs to be called synchronously when using the node
+      // plugin as otherwise it would exit prematurely.
+      // This should be fixed so it can be used in line with the other cases.
       next();
     } else if (opts.wd) {
       withServer(b, opts, function () {
--- a/lib/mochify.js
+++ b/lib/mochify.js
@@ -41,7 +41,9 @@ function withServer(b, opts, callback) {
       callback();
     });
   } else {
-    callback();
+    // ensure the callback is always called asynchronously, no matter
+    // if a server is created or not
+    process.nextTick(callback);
   }
 }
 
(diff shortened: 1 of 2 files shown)


m90:
Let me look into why Node 6 is failing first ...

m90:
Looks like the Travis failure is also happening on a fresh branch off `master`: https://travis-ci.org/mantoni/mochify.js/jobs/413987394

If you force Node 6 to use npm@6 though it passes again: https://travis-ci.org/mantoni/mochify.js/builds/413989994 which tells me we have some #171 situation once more that is not related to this PR/issue combo and will need a separate fix.

mantoni:
I reccon we can partially avoid these issues by using npm@6 and install with `npm ci` to make sure we get exacyly what we have locally.

mantoni:
Please rebase against master. That should fix the build.

lib/mochify.js
```diff
@@ -154,51 +156,60 @@ module.exports = function (_, opts) {
     noColors        : opts.noColors
   };
 
+  (function initPlugins(next) {
+    if (opts.consolify) {
+      b.plugin(consolify, opts);
+      next();
+    } else if (opts.node) {
+      b.plugin(node, opts);
+      next();
```
m90: This is one thing I still don't understand: in theory this should be `process.nextTick(next)` as well, so that the execution order of the blocks is always the same:

- initPlugins
- main body + return
- appendAdditionalPlugins

Yet, when I do this, the `node/plugin` tests will start failing as it looks as if the tests all run synchronously before `appendAdditionalPlugins` is even called.

I'm wondering if this is a pointer to a more fundamental problem here?

m90:
@mantoni I rebased and the build is green again. I left one note in the diff though.

lib/mochify.js
m90: A possible way to solve this could look like this: https://github.com/mantoni/mochify.js/commit/cd4252fe5ef1ac0c47da2a3eae3914b3ed247864

mantoni: I ran a few checks and I presume there seems to be a slightly bigger issue with the `node` runner here. It's adding a through stream to the browserify pipeline to capture the input `js` and when changing the line here to `process.nextTick(next);` that stream ends immediately causing the node plugin to abort.

The way I'm testing these things is by adding `console.error(stderr);` in `test/fixture/run.js` and then put `console.error('!');` statements in the code. I don't have the capacity to fix this right now.

mantoni: We'd be sacrificing quite a bit of performance when delaying the `bundle` call until everything else is running. Browserify will only run the transforms and plugins on the code once `bundle()` is called which is currently happening while chrome starts up, the web server starts up and chrome loads the URL in a tab. The code is currently written to do as much setup concurrently as possible.

mantoni:
I'd say if this fixes the issue at hand, it's good to merge. We can adress the node plugin foo with async bundles separately.

m90:
I'll add a comment on why the callback is called async or sync so noone forgets and then we'll have `5.8.1`, should we?

PR #180, opened 2018-08-10
Do not add plugins on browserify object asynchronously to ensure predictable order

m90:
I looked into the `node` plugin issue from #178 which led nowhere and gave me a terrible headache :pill:. I feel browserify does not really expect plugins and transforms to be "maybe" added asynchronously, at least I wasn't able to find any logic in what was called in which order if plugins were added like this.

Instead I moved all the main action on `b` back into the synchronous `mochify` call, so it's 100% foolproof that the calls will be executed in the supposed order.

The server will still be created asynchronously, which means we rely on puppeteer to wait for it to be ready, but that's what's happening at the moment as well and is the price for not deferring `bundle()` until setup has happened.

I feel this is quite a bit saner than #178, but maybe there's other input as well.

### DIFF
--- a/lib/mochify.js
+++ b/lib/mochify.js
@@ -24,26 +24,9 @@ var trace      = require('./trace');
 var args       = require('./args');
 var server     = require('./server');
 
-function withServer(b, opts, callback) {
-  var serverPort = opts['https-server'];
-  if (Number.isInteger(serverPort)) {
-    server(b, serverPort, opts.watch, function (err, actualPort) {
-      if (err) {
-        console.error('Error: Failed to launch server\n' + err.stack + '\n');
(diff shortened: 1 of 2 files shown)


mantoni:
I agree with your strategy. Feel free to merge 👍

mantoni:
Hm. Unfortunately the build is now failing on master on node 6 only. But everything we merged was green 🤔

m90:
Isn't the failing build on a different branch (one that I used solely to understand why the one test is being skipped)? Or does master fail too now?

mantoni:
Hmm, the badge in the readme was red. But false alarm, apparently. It’s green again ¯\_(ツ)_/¯

PR #181, opened 2018-08-11
Consolidate validation of options in one place

m90:
When looking at #180 I noticed that the validation of the server port was done in the `args` module, when everything else would be validated in `mochify`. This simply moves things to the same location.


mantoni:
Sorry to keep this open for so long. I didn't pay attention 😄
Looks fine, thank you!

PR #182, opened 2018-08-11
Update browserify to v16

m90:
And while we're at it, Browserify could also be bumped to v16, I think mochify does not require a major bump to do so until now?


mantoni:
Usually I do majors if either Mocha or Browserify majors are updated. Mocha is currently also a version behind.

m90:
Ah ok, I checked [v5.0.0](https://github.com/mantoni/mochify.js/blob/v5.0.0/package.json) which was on Browserify v14 which made me think it's handled like this, but then let's not do this until there's a real reason to do so, or there's something else that would require a major and it could sneak in.

PR #185, opened 2018-11-18
nyc support

mantoni:
~~_Work in progress! Do not merge!_~~

This ~~is an attempt to integrate~~ integrates mochify with `nyc`. It works ~~for small projects~~, but requires quite some complicated setup.

It might be a better option to write a custom transform in this project, that uses the recent implementation of the istanbul instrumenter directly.

Ref https://github.com/sinonjs/sinon/issues/1948

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ the test framework.
     - Supports [BrowserStack][] ([docs](#browserstack-setup))
     - Supports [WebDriver][] ([docs](#selenium-webdriver-setup))
 - Code coverage options:
+    - Using [nyc][] ([docs](#code-coverage-with-nyc))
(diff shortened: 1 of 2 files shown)


mantoni:
It's quite a hack, but it works fine. I'd go with this for now. I've applied the new instructions in the readme successfully to a small project and to Sinon.JS.

mantoni:
I tried coming up with a test case, but it's really hard, given that the tests are all integration style.

mroderick:
That is indeed an elaborate script

mantoni:
I managed to bring it down by using a nyc config in the `package.json` and by invoking `npm run test-headless` decorated with the additional argument. I just released mochify 6 with all the updates and will send a PR for Sinon now.

PR #186, opened 2018-11-18
Mocha 5.2, Puppeteer 1.10 and Studio Changes 1.7

mantoni:
Update the core dependencies to more recent versions.

New `changes --commits --footer` features generate a more expressive changelog.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "test": "mocha",
     "posttest": "npm run lint",
     "preversion": "npm test",
-    "version": "changes",
+    "version": "changes --commits --footer",
     "postversion": "git push --follow-tags && npm publish"
   },
   "repository": {
@@ -82,7 +82,7 @@
     "watchify": "^3.11.0"
   },
   "devDependencies": {
-    "@studio/changes": "^1.4.0",
+    "@studio/changes": "^1.7.0",
     "eslint": "^3.19.0",
     "tmp": "0.0.28"
   },
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "min-wd": "^2.11.0",
     "mocaccino": "^3.0.0",
     "mocha": "^5.2.0",
-    "puppeteer": "^1.6.2",
+    "puppeteer": "^1.10.0",
     "resolve": "^1.5.0",
     "source-mapper": "^2.1.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -73,16 +73,16 @@
     "glob": "^7.1.2",
     "min-wd": "^2.11.0",
     "mocaccino": "^3.0.0",
-    "mocha": "^4.1.0",
-    "puppeteer": "^1.6.2",
+    "mocha": "^5.2.0",
```
m90: Does this (together with [`browserify` now at 16](https://github.com/mantoni/mochify.js/commit/1bf51a26911de02e44a61b089676b5705b4ea0d6#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R70)) mean we'll have `mochify@6`?

mantoni: I was asking this myself, but I think a minior bump is fine since this does not introduce breaking changes in mochify. I changed the changelogs of these projects and the only breaking change that is theoretically affecting mochify could be mocha dropping support for IE 9 and IE 10. Most likely, this does not even break in the context of mochify with min-webdriver. If it does and there is somebody using this combination, they'd have to nail their dependency to an older version. Or shall we play it safe and bump the major?

m90: The mocha changelog even states the following regarding IE9 and IE10 support:

> Practically speaking, only code which consumes (through bundling or otherwise) the userland buffer module should be affected.

so summing this up with the min-webdriver situation I would be surprised if someone gets bitten by this, and if it really happens people can always lock down to the current version. Also, [v5.0.0](https://github.com/mantoni/mochify.js/blob/v5.0.0/package.json#L67) was on browserify 14 still, so some minor version in the v5 line must have already done a major dependency update.

I'd be fine with making this v5.9.0

mantoni: It's quite a few moving parts, with a few bumps in Chromium and all. I made up my mind and I'm going to release this as a major. Version numbers are free, after all. But it's good to know that it should be a frictionless upgrade in 99.95% of the cases.

mantoni:
This was released in v6.0.0. [A follow-up patch to v6.0.1](https://github.com/mantoni/mochify.js/pull/187) upgraded Moccachino to v4 in order to actually upgrade to Mocha 5 🙈

PR #187, opened 2018-11-19
Upgrade Mocaccino to v4

mantoni:
Mocaccino used to load Mocha 4 still, causing two Mocha versions to be installed.

This is going to be a patch for v6.


mantoni:
Release in v6.0.1.

PR #193, opened 2018-11-25
Improve test cases for failure cases

m90:
I found out something interesting about the test case added in #192 : it would also pass against the previous (broken) version :confused: 

Which happened due to the following constellation: the last test in the test (which runs after the failing async one) in here: https://github.com/mantoni/mochify.js/blob/1de1c0010e0035a0e762a208f549711cc97d05ec/test/fixture/fails-async/test/fails-async.js#L4-L21 is synchronous. This means that it would still be executed (alongside all of the test suite teardown) *before* the exit is triggered as the forceful exit in the mochify command is being deferred until next tick: https://github.com/mantoni/mochify.js/blob/1de1c0010e0035a0e762a208f549711cc97d05ec/bin/cmd.js#L19-L25

This means in order to fully recreate the condition that was previously breaking, we need to add another async test *after* the failing one, so that we can check if mochify exits prematurely or not.

The tests like this will now fail on `6.0.3` and pass on `6.0.4`.


mantoni:
Interesting case. Since that’s not entirely obvious when reading the test case, a comment would help to remember why this test was written. Another question is if we can merge both tests into a single asynchronous test?

m90:
> Another question is if we can merge both tests into a single asynchronous test?

I'm only 99% sure I get this. Are you talking about merging `fails` and `fails-async`? In case yes, that would make sense, yeah.

m90:
As I will be afk for a little while now, I added a comment and merged the two test cases I think you were referring to. I also added an additional check to the test for immediate exit on evaluation error (which is the behavior we want to keep after #192).

Feel free to merge this as is if you like it, or leave feedback and I can address it mid next week.

test/node-test.js
```diff
@@ -29,43 +29,35 @@ describe('node', function () {
...
+      assert.equal(lines[12], '# fail 2');
       done();
     });
   });
 
-  it('handles async failures', function (done) {
-    run('fails-async', ['--node', '-R', 'tap'], function (code, stdout) {
+  it('fails and exits if `describe` throws', function (done) {
+    run('describe-throws', ['--node', '-R', 'tap'], function (code, stdout) {
+      assert.equal(stdout.indexOf('# node:'), 0);
```
m90: Somehow this is all that node prints in this situation, I have to admit I don't know who is eating the error message here?

mantoni:
👍  Thank you!

PR #195, opened 2019-02-27
add support for custom mocha path

boneskull:
Needs changes from mantoni/mocaccino.js#23... more discussion there.


mantoni:
Thank you! Should be easy to integrate with mocaccino `v4.1.0`. 👍

boneskull:
@mantoni 	not sure if the failure is related..

mantoni:
Apparently it wasn't 😄

mantoni:
Can you add an integration test for this?

m90:
This `coverage dot` test is the bane of mochify's test suite, maybe it would be a thing to find out why it is so flaky.

boneskull:
dot reporter tests have historically been a bane on mocha's side as well

boneskull:
we have moved towards using the JSON reporter for most tests where this level of information is needed

boneskull:
awww who clicked the restart job button

boneskull:
OK, I've added two tests, and should be done w/ this assuming CI passes.

mantoni:
Released in `v6.1.0` 🎈

PR #197, opened 2019-03-13
Inspect page error objects

mantoni:
This is an attempt to get more information than `[object Object]` in the failing build here: https://github.com/sinonjs/sinon/pull/1995

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "min-wd": "^2.11.0",
     "mocaccino": "^4.1.0",
     "mocha": "^5.2.0",
-    "puppeteer": "^1.10.0",
+    "puppeteer": "^1.13.0",
     "resolve": "^1.5.0",
     "source-mapper": "^2.1.0",
     "subarg": "^1.0.0",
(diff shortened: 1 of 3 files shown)


mantoni:
Released in `v6.1.1`.

package.json
```diff
@@ -75,12 +75,12 @@
     "min-wd": "^2.11.0",
     "mocaccino": "^4.1.0",
     "mocha": "^5.2.0",
-    "puppeteer": "^1.10.0",
+    "puppeteer": "^1.13.0",
```
m90: Looking at the changelog for this: https://github.com/GoogleChrome/puppeteer/releases I've seen that puppeteer now has experimental Firefox support. Is this something mochify should/could be supporting too?

mantoni: Absolutely!

m90: I just tried hot-swapping and it seems like it's not entirely working like the Chrome version as `page.on('console'` will never be called (and all `file://` based tests time out) which looks to be caused by this issue: https://github.com/GoogleChrome/puppeteer/issues/3652

I guess I'll just open an issue for adding the option once the upstream bug has been resolved.

PR #200, opened 2019-03-20
Enable support for passing --dumpio to Puppeteer

### LINKED ISSUE #199: Support `--dumpio` option for tracing down errors with chromium
As described in https://github.com/sinonjs/sinon/pull/1999, it's quite hard to find issues with Chromium since there is no output. The `dumpio: true` option should be supported by mochify through a command line switch.

fatso83:
To make debugging easier, support passing this option on to puppeteer.

With regards to the test, I am not at all sure it's what you want. I just wasn't able to figure out how to do a regex match that would produce meaningful assertion errors using the standard `assert` module, so I decided just to test on a subset of the string.

closes #199 

Background: sinonjs/sinon#1999

### DIFF
--- a/lib/args.js
+++ b/lib/args.js
@@ -15,6 +15,7 @@ var defaults = {
   cover: false,
   node: false,
   debug: false,
+  dumpio: false,
   wd: false,
   recursive: false,
   'ignore-ssl-errors': false,
@@ -36,7 +37,7 @@ function args(argv) {
       'mocha-path'],
     boolean: ['help', 'version', 'watch', 'cover', 'node', 'wd', 'debug',
       'invert', 'recursive', 'colors', 'ignore-ssl-errors', 'browser-field',
-      'commondir', 'allow-chrome-as-root', 'async-polling'],
+      'commondir', 'allow-chrome-as-root', 'async-polling', 'dumpio'],
     alias: {
       help: 'h',
       version: 'v',
--- a/test/chromium-test.js
+++ b/test/chromium-test.js
@@ -121,6 +121,62 @@ describe('chromium', function () {
     });
   });
 
+  it('outputs a minimal error message to stderr when --dumpio is not specified',
+    function (done) {
+      run('fails', ['-R', 'dot', '--no-colors'],
+        function (code, stdout, stderr) {
+          assert.equal(stderr, 'Error: Exit 1\n\n');
+          assert.equal(code, 1);
+          done();
+        });
+    }
+  );
+
+  it('receives all console errors on stderr when --dumpio is given',
+    function (done) {
+      run('fails', ['-R', 'dot', '--no-colors', '--dumpio'],
+        function (code, stdout, stderr) {
+
+          var stdoutLines = stdout.split('\n').slice(1); // skip 'chromium:'
+
+          /* eslint-disable max-len */
+          // The sub-set lines actually relating to the console output.
+          // Other lines may relate to internal Chrome errors, such as
+          // '[0322/162300.874805:ERROR:command_buffer_proxy_impl.cc(125)] ContextResult::kTransientFailure: Failed to send GpuChannelMsg_CreateCommandBuffer.'
+          var stderrLines = stderr
+            .split('\n')
+            .filter(function (l) { return l.indexOf('INFO:CONSOLE') >= 0; });
+
+          // What actually ends up on the output is quite unstable:
+          // Ref https://gist.github.com/fatso83/2bb8c4ff82fa66f66db5030b0af06d66
+          // Sometimes you get a stack dump, sometimes not.
+          // We just check the portion that actually seems to always be present
+          /* eslint-enable */
+          var numberOfLinesToCheck = 3;
+
+          for (var index = 0; index < numberOfLinesToCheck; index++) {
+            var stdoutLine = stdoutLines[index];
+            var stderrLine = stderrLines[index];
+            var assertionError = 'Should have matched: \''
+              + stdoutLine
+              + '\' with \''
+              + stderrLine
+              + '\'';
+
+            assert(stderrLine.match(stdoutLine), assertionError);
+            assert(
+              stderrLine.match(/^\[.*:INFO:CONSOLE\(\d+\)\] "/),
+              assertionError);
+            assert(
+              stderrLine.match(/source: __puppeteer_evaluation_script__/),
+              assertionError);
+          }
+
+          assert.equal(code, 1);
+          done();
+        });
+    });
+
   it('uses custom chrome', function (done) {
     run('passes', ['--chrome', 'some/path'], function (code, stdout, stderr) {
       assert.equal(stdout, '');
--- a/lib/chromium.js
+++ b/lib/chromium.js
@@ -41,6 +41,7 @@ module.exports = function (b, opts) {
 
   var options = {
     devtools: !!opts.debug,
+    dumpio: !!opts.dumpio,
     ignoreHTTPSErrors: opts['ignore-ssl-errors'],
     args: ['--allow-insecure-localhost']
   };
@@ -52,6 +53,7 @@ module.exports = function (b, opts) {
   if (opts.chrome) {
     options.executablePath = opts.chrome;
   }
+
 
   function finish() {
     if (output) {
--- a/test/args-test.js
+++ b/test/args-test.js
@@ -135,6 +135,12 @@ describe('args', function () {
     assert.equal(opts.chrome, '/foo/bar');
   });
 
+  it('parses --dumpio', function () {
+    var opts = args(['--dumpio']);
+
+    assert(opts.dumpio);
+  });
+
   it('parses --ignore-ssl-errors', function () {
     var opts = args(['--ignore-ssl-errors']);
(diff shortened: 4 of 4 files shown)


test/chromium-test.js
```diff
@@ -121,6 +121,32 @@ describe('chromium', function () {
...
+        + '  1) test\n'
+        + '       fails synchronously:\n'
+        + '     Error: Oh noes!\n'
+        + '      at Context.<anonymous> (test/fails.js:6)\n'
+        + '\n'
+        + '  2) test\n'
+        + '       fails asynchronously:\n'
+        + '     Uncaught Error: Script error. (:0)';
+
+        assert.equal(stdout.slice(0, 40), expected.slice(0, 40));
```
mantoni: I don't understand this assertion. The `stdout` value does not seem to be relevant for this test.

lib/args.js
```diff
@@ -36,7 +37,7 @@ function args(argv) {
       'mocha-path'],
     boolean: ['help', 'version', 'watch', 'cover', 'node', 'wd', 'debug',
       'invert', 'recursive', 'colors', 'ignore-ssl-errors', 'browser-field',
-      'commondir', 'allow-chrome-as-root', 'async-polling'],
+      'commondir', 'allow-chrome-as-root', 'async-polling', 'dumpio'],
```
m90: Should this print a warning (or even fail) when being passed together with `--node`?

mantoni: I think we don't fail on invalid argument combinations at all yet, do we?

m90: I think we do here: https://github.com/mantoni/mochify.js/blob/90886285fea8ec819ce622861976c57bd70103ad/lib/mochify.js#L70-L79 ?

mantoni: Yeah, right. But in general, there are heaps of invalid / useless combinations, like, all chromium specific options with `--wd` or `--consolify` ... If we want to do this, I'd say it's a separate change introducing some sort of valid option mapping.

m90: Yeah, I agree. Let's not do this here.

test/chromium-test.js
fatso83: Pardon for being unclear - I could have explained this in a comment of some sort:

Basically, there is no output if `--dumpio` is not given. A failing test just results in `''` - the empty string. So this is basically just a way of asserting that the output is produced at all.

If I remove the `--dumpio` option the test fails with this:

```javascript
      Uncaught AssertionError [ERR_ASSERTION]: '' == '# chromium:\n\n\nError: Oh noes!\n      at t'
      + expected - actual

      +# chromium:
      +
      +
      +Error: Oh noes!
      +      at t
```

fatso83: I couldn't come up with any other way of asserting that this worked, but I take suggestions :smiley_cat:

mantoni: Hmmm, I guess I have to play with this myself to understand. Seems dodgy that regular test output is swallowed if this flag is not given. It should always do that, IMO.

mantoni: So, I check out you code and found that it makes no difference if I'm removing `--dumpio`, the test is still green. The flag causes puppeteer to write a lot of detailed information to `stderr`, which is passed back to the test as a third argument. I guess that the test should assert.

I also ran tests on other projects with this flag and since it dumps _all_ IO, the test output is shown twice. So it would make sense not print anything to standard out ourselves if `--dumpio` is given. Makes sense?

fatso83: Oh, man. I have no idea how that happened. I was *so sure* I had tested with/without the dumpio option and saw different behaviour. It must be because I worked from morning until afternoon without eating. That obviously just earned me an award in the league of morons  :1st_place_medal:  

Fixing it now, but I am still not sure how to do a proper part-wise match on the output. Unless you get a better suggestion, I'll stick to slicing on sub-parts of the output, which feels less than great ...

So for the following output, I'd match on `DevTools listening on ws://127.0.0.1`, `[.*:INFO:CONSOLE(.*)]`, `source: __puppeteer_evaluation_script__` and the output from stdout.
```
DevTools listening on ws://127.0.0.1:37763/devtools/browser/90ab4733-514e-445b-b88a-97915efbe68d
[0322/153940.105525:ERROR:command_buffer_proxy_impl.cc(106)] ContextResult::kTransientFailure: Shared memory region is not valid
[0322/153940.333397:INFO:CONSOLE(1415)] "", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.337272:INFO:CONSOLE(1415)] "", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.391361:INFO:CONSOLE(1347)] "Uncaught Error: Oh noes!", source: __puppeteer_evaluation_script__ (1347)
[0322/153940.450332:INFO:CONSOLE(1415)] "  !!..", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.450540:INFO:CONSOLE(1415)] "", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.450926:INFO:CONSOLE(1415)] "  2 passing (119ms)", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.451132:INFO:CONSOLE(1415)] "  2 failing", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.451323:INFO:CONSOLE(1415)] "", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.451818:INFO:CONSOLE(1415)] "  1) test", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.452002:INFO:CONSOLE(1415)] "       fails synchronously:", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.452168:INFO:CONSOLE(1415)] "     Error: Oh noes!", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.452339:INFO:CONSOLE(1415)] "      at Context.<anonymous> (__puppeteer_evaluation_script__:1341:11)", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.452476:INFO:CONSOLE(1415)] "      at callFn (__puppeteer_evaluation_script__:6032:21)", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.452617:INFO:CONSOLE(1415)] "      at Test.Runnable.run (__puppeteer_evaluation_script__:6024:7)", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.452857:INFO:CONSOLE(1415)] "      at Runner.runTest (__puppeteer_evaluation_script__:6560:10)", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.452942:INFO:CONSOLE(1415)] "      at __puppeteer_evaluation_script__:6678:12", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.453063:INFO:CONSOLE(1415)] "      at next (__puppeteer_evaluation_script__:6474:14)", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.453217:INFO:CONSOLE(1415)] "      at __puppeteer_evaluation_script__:6484:7", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.453382:INFO:CONSOLE(1415)] "      at next (__puppeteer_evaluation_script__:6408:14)", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.453476:INFO:CONSOLE(1415)] "      at __puppeteer_evaluation_script__:6452:5", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.453701:INFO:CONSOLE(1415)] "      at timeslice (__puppeteer_evaluation_script__:1564:27)", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.453781:INFO:CONSOLE(1415)] "", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.453872:INFO:CONSOLE(1415)] "  2) test", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.454070:INFO:CONSOLE(1415)] "       fails asynchronously:", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.454141:INFO:CONSOLE(1415)] "     Uncaught Error: Script error. (:0)", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.454292:INFO:CONSOLE(1415)] "      at global.onerror (__puppeteer_evaluation_script__:1546:10)", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.454414:INFO:CONSOLE(1415)] "", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.454499:INFO:CONSOLE(1415)] "", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.454667:INFO:CONSOLE(1415)] "", source: __puppeteer_evaluation_script__ (1415)
[0322/153940.455152:INFO:CONSOLE(12)] "[EXIT 1]", source: __puppeteer_evaluation_script__ (12)
```

mantoni: Don't be so hard on yourself 😉 I guess it doesn't even have to be the entire output. It should be good enough to just find a line or some part of the output that is unique, like, checking if `__puppeterr_evaluation_source` exists in the string.

fatso83: It's fixed. Check again.

mantoni:
Released in `v6.2.0`. The `--dumpio` option is now available. Thanks everyone 💯

PR #201, opened 2019-03-23
Dumpio part 2

mantoni:
See discussion on https://github.com/mantoni/mochify.js/pull/200.

### DIFF
--- a/lib/chromium.js
+++ b/lib/chromium.js
@@ -135,6 +135,8 @@ module.exports = function (b, opts) {
           return;
         }
         if (!options.dumpio) {
+          // Only print console output if `--dumpio` is not specified, or the
+          // output is printed twice.
           output.write(sourceMapper.line(consumer, text));
           output.write('\n');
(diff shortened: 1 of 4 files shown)


mantoni:
Sorry for the noise. I could only experiment with the build on Travis since the output is different from my local setup. The build should be good now.

lib/chromium.js
```diff
@@ -134,8 +134,10 @@ module.exports = function (b, opts) {
           pageFinish();
           return;
         }
-        output.write(sourceMapper.line(consumer, text));
-        output.write('\n');
+        if (!options.dumpio) {
```
m90: I'm wondering if this behaviour (nothing but the header written to stdout) should be mentioned in the docs somehow, as I found it a little surprising when reading the test?

mantoni: I added an inline comment to explain that this is here to avoid printing everything twice when the option is enabled.

PR #71, opened 2015-01-26
Add `--web-security` option to toggle PhantomJS web security

krawaller:
As discussed in https://github.com/mantoni/mochify.js/issues/69.

Please let me know if you want me to tweak it somehow!

PR #203, opened 2019-06-06
[feat] Bring back --web-security flag

decompil3d:
There used to be a flag in old versions called `--web-security` (added in #71) which allowed disabling the same-origin policy in Phantom. Now that we use Puppeteer/Chromium, that flag was removed. This PR adds it back in with the exact same semantics it used to have, but it now passes through to Chromium.

Happy to adjust as needed. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,7 @@ break at the `debugger` statement.
 - `--help` or `-h` shows usage and all available options.
 - `--async-polling` disables async polling when set to false (for use in Appium).
 - `--mocha-path` specifies path to a custom Mocha module
+- `--web-security` allows disabling same-origin policy (when set to `false`, passes down `--disable-web-security` to Chromium)
 
 ## Continuous Integration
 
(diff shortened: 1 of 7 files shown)


mantoni:
Thanks a lot! Nice to bring this feature back. While I like how you wrote the integration test, I'm not a fan of making outbound requests in test cases. I think it should be possible to make a local test by creating a tiny server that sets the `Access-Control-Allow-Origin` header and making a localhost request. What do you think?

decompil3d:
@mantoni Good idea. Fixed in 41559b9.

decompil3d:
I didn't have to set the ACAO header since, by default, the tests are setting an origin of `null`, so it will still fail (as expected) with a CORS error even without that header.

decompil3d:
@mantoni Unfortunately, the "Coverage Dot" and "Coverage Tap" tests are really flaky. I can re-run quite a bit to get all 3 Node versions to pass. Seems like other PRs before mine have also hit this.

mantoni:
Yes, sorry about the flaky tests. I'm struggeling to identify the issue. Seems like some Chromium internal timings that I'm not accounting for.

Thanks for adjusting the test. Looks good to me 👍

mantoni:
📦 Released in `v.6.3.0`. I've also managed to address the build stability – hopefully it's bothering us less now. See [release notes](https://github.com/mantoni/mochify.js/blob/master/CHANGES.md) for details. Thank you for contributing ❤️

PR #205, opened 2019-07-23
Puppeteer 1.19.0

mantoni:
This PR seems to finally fix the flaky builds. I ran the tests several times locally and the Travis build on the branch is also green on the first attempt 🎉

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "min-wd": "^2.11.0",
     "mocaccino": "^4.1.0",
     "mocha": "^5.2.0",
-    "puppeteer": "^1.13.0",
+    "puppeteer": "^1.19.0",
     "resolve": "^1.5.0",
     "source-mapper": "^2.1.0",
     "subarg": "^1.0.0",
--- a/test/chromium-test.js
+++ b/test/chromium-test.js
@@ -446,37 +446,36 @@ describe('chromium', function () {
 
   context('https-server with a port value given', function () {
     var server;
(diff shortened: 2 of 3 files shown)


test/chromium-test.js
```diff
@@ -446,37 +446,36 @@ describe('chromium', function () {
...
+          port = String(server.address().port);
+        }
         done(err);
       });
     });
 
-    after(function () {
-      server.close();
+    after(function (done) {
+      server.close(done);
```
m90: Does this mean the flaky tests were a race condition between a server starting and another one potentially not being closed yet?

package.json
```diff
@@ -75,7 +75,7 @@
     "min-wd": "^2.11.0",
     "mocaccino": "^4.1.0",
     "mocha": "^5.2.0",
-    "puppeteer": "^1.13.0",
+    "puppeteer": "^1.19.0",
```
m90: Puppeteer changelog looks unsuspicious :+1:

test/chromium-test.js
mantoni: Yes. At least after upgrading puppeteer that was the only randomly failing issue. We had other issues before as well.

package.json
mantoni: Thanks for looking through. I didn’t wait with merging because master was failing again and I couldn’t get it green and this change worked in every environment all the time ¯\_(ツ)_/¯

PR #209, opened 2019-10-09
Add support for the browserify `--no-detect-globals` option

mantoni:
By default, browserify detects the use of global like `process` and `global`. This is sometimes undesireable and recently caused issues with upgrading browserify in the Sinon projects (see https://github.com/sinonjs/commons/pull/37, https://github.com/sinonjs/sinon/pull/2119 and most importantly the analysis from @fatso83 in https://github.com/sinonjs/sinon/issues/2085).

Allowing to turn off this browserify behavior brought a chain of other issues to the light:

- The `brout` project (used internally) assumed that `process` is shimed. This has been addressed in `v1.3.0` (see https://github.com/mantoni/brout.js/commit/9fc88568629256cf47f03e8744124f50e4651a64).
- The `mocaccino` project (also used internally) assumed that `process.nextTick` is shimed. This has been addressed in `v4.1.2` (see https://github.com/mantoni/mocaccino.js/commit/f46487a16d1da04bcad60d111ea98b900e2b62e4).

Unfortunately, the `process` object _must_ be installed by `brout`, or most Mocha reporters would stop working. However, it's a good improvement to not have `global` and things like `process.nextTick` shimed by browserify in our test runs.

I have tested these changes with the Sinon project by adding the switch. This highlights several issues with feature detections, mostly in the test cases, but also in `lolex`. Once this change has been released, I'll send PRs with fixes for those projects.

### DIFF
--- a/test/chromium-test.js
+++ b/test/chromium-test.js
@@ -463,6 +463,24 @@ describe('chromium', function () {
       });
   });
 
+  it('detects globals by default', function (done) {
+    run('detect-globals', ['-R', 'tap'],
+      function (code, stdout, stderr) {
+        assert.equal(stderr, 'object\n');
+        assert.equal(code, 0);
+        done();
+      });
+  });
+
+  it('allows to turn off global detection', function (done) {
+    run('detect-globals', ['--no-detect-globals'],
+      function (code, stdout, stderr) {
+        assert.equal(stderr, 'undefined\n');
+        assert.equal(code, 0);
+        done();
+      });
+  });
+
   context('https-server with a port value given', function () {
     var server;
     var port;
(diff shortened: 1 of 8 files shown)


test/chromium-test.js
```diff
@@ -463,6 +463,34 @@ describe('chromium', function () {
...
+        assert.equal(stderr, 'object\n');
+        assert.equal(code, 0);
+        done();
+      });
+  });
+
+  it('allows to turn off global detection', function (done) {
+    run('detect-globals', ['--no-detect-globals'],
+      function (code, stdout, stderr) {
+        assert.equal(stderr, 'undefined\n');
```
fatso83: nice!

boneskull:
@mantoni did you need specific feedback on something here

mantoni:
No, I just thought you might have ideas or additional input on the topic.

PR #216, opened 2020-11-09
Upgrade ESLint to latest

mroderick:
And fix no-prototype-builtins errors

See: https://eslint.org/docs/rules/no-prototype-builtins

#### Background

In another project I am trying to see if I can upgrade away from old `minimist` versions, to not have any security warnings.

Specifically https://www.npmjs.com/advisories/1179 against `minimist`.

The old version of `eslint` would cause an affected version of `minimist` to be installed.
Upgrading `eslint` to latest is safe, and takes care of one of the many versions of `minimist` installed, making my task a bit easier.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-  - "6"
-  - "8"
   - "10"
   - "12"
+  - "14"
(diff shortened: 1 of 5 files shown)


mantoni:
The build is failing because the new eslint version uses language features that are not available on node 6 and 8. I think it's overdue to drop support for those node versions anyway and add a build for node 14.

mroderick:
@mantoni I agree. I've pushed fresh commits and it's all green

mantoni:
Released in `v7.0.0` (https://github.com/mantoni/mochify.js/blob/master/CHANGES.md#700)

PR #221, opened 2021-05-16
Support nyc v15

albertyw:
From what I can tell, the only thing that breaks with nyc v15 is that it removes `NYC_ROOT_ID` (https://github.com/istanbuljs/nyc/blob/master/CHANGELOG.md#-breaking-changes) which is depended on by mochify.  I'm not sure how the ID is used, but it seems like any ID will work.  Therefore, for nyc>=15, I'm generating a random `NYC_ROOT_ID`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,7 +330,7 @@ The `mochify` function returns a Browserify instance. Please refer to the
 
 ## Code coverage with NYC
 
-Install `nyc@14` (v15 is not yet supported!), the `babelify` transform,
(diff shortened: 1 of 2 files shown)


albertyw:
@m90 @mantoni Can I get a review/merge on this?

mantoni:
Thank you 💯 for fixing this, and sorry to keep you waiting.

📦 Released in `v7.1.1`.

PR #225, opened 2021-06-03
Migrate CI to GitHub actions

m90:
The __org__ version of Travis seems to be shutting down on the 15th of June which isn't that far into the future anymore.

It would be possible to migrate the setup to the __com__ version (https://docs.travis-ci.com/user/migrate/open-source-repository-migration), but tbh that feels like a dead end. Instead, Mochify can use GitHub actions I guess.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,28 @@
+name: Build
+
+on:
+  push:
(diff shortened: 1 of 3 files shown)


mantoni:
Great! Thank you 🙏

Issue #224, opened 2021-05-26
Use Browserify 17


m90:
Ideally, Mochify would use Browserify 17 (instead of 16 at the time of writing) to keep in sync with maintained versions of the underlying tools.

---

Picking up the comment here: https://github.com/mantoni/mochify.js/pull/223#issuecomment-847258384 I looked into why such an update from 16 to 17 is breaking Mochify and found out the following (this issue is mostly about collecting what I found out whenever this updates becomes required for some reason):

1. There is yet another new deprecation warning about the usage of `SharedArrayBuffer` that litters `stderr`, but that can be swallowed just as Mochify already does with other deprecation messages. (https://github.com/browserify/node-util/issues/58)
1. After muting these warnings, the tests for `--no-detect-globals` fail because of the following:
Browserify 17 updates `util` to 0.12 https://github.com/browserify/browserify/pull/1844, containing this commit https://github.com/browserify/node-util/commit/6b828255a7f407efcd7e4d2c54ddb43256e491fb which is reading `process.env.NODE_DEBUG` without checking whether `process` is even defined. This means that when passing the option not to shim globals to browserify, any code that requires `util` (Mocha for example) will fail with an `Uncaught ReferenceError: process is not defined`

I'm not entirely sure if this is something that could be or should be worked around in Mochify or if this is something that is to be considered an upstream bug in either `node-util` or `browserify`. Mochify could of course also drop the `--no-detect-globals` flag again but this would mean https://github.com/mantoni/mochify.js/issues/208 comes back to life.

m90:
There's an issue in `node-util` about this https://github.com/browserify/node-util/issues/43

PR #226, opened 2021-06-14
Update watchify package to v4.0.0

albertyw:
`watchify` v3 has indirect security vulnerabilities.  `watchify` v4 resolves those vulnerabilities and has minimal incompatibilities.

[Watchify changelog](https://github.com/browserify/watchify/blob/master/CHANGELOG.md)

~~Also updates `puppeteer` to fix builds with node v16.  [Puppeteer changelog](https://github.com/puppeteer/puppeteer/blob/main/CHANGELOG.md)~~

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,7 @@
         "source-mapper": "^2.1.0",
         "subarg": "^1.0.0",
         "through2": "^2.0.0",
-        "watchify": "^3.11.1"
+        "watchify": "^4.0.0"
       },
       "bin": {
         "mochify": "bin/cmd.js"
@@ -350,12 +350,15 @@
       }
     },
     "node_modules/anymatch": {
-      "version": "2.0.0",
-      "resolved": "https://registry.npmjs.org/anymatch/-/anymatch-2.0.0.tgz",
-      "integrity": "sha512-5teOsQWABXHHBFP9y3skS5P3d/WfWXpv3FUpy+LorMrNYaT9pI4oLMQX7jzQ2KklNpGpWHzdCXTDT2Y3XGlZBw==",
+      "version": "3.1.2",
+      "resolved": "https://registry.npmjs.org/anymatch/-/anymatch-3.1.2.tgz",
+      "integrity": "sha512-P43ePfOAIupkguHUycrc4qJ9kz8ZiuOUijaETwX7THt0Y/GNK7v0aa8rY816xWjZ7rJdA5XdMcpVFTKMq+RvWg==",
       "dependencies": {
-        "micromatch": "^3.1.4",
-        "normalize-path": "^2.1.1"
+        "normalize-path": "^3.0.0",
+        "picomatch": "^2.0.4"
+      },
+      "engines": {
+        "node": ">= 8"
       }
     },
     "node_modules/argparse": {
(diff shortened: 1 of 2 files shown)


m90:
Thank you for this PR.

> Also updates puppeteer to fix builds with node v16.

The last build passed in Node 16 using Puppeteer 9. Could you elaborate on what exactly is breaking here? Updating puppeteer in Mochify can introduce very subtle breakages from time to time...

albertyw:
Did puppeteer 9 build correctly?  From https://github.com/mantoni/mochify.js/actions, build 6 and 7 were only with the watchify update and both failed.  build 8 included the puppeteer update and succeeded.  A weird thing I noticed for both build 6 and 7 was that 
 - on `npm ci`, there's no `Chromium (869685) downloaded to /home/runner/work/mochify.js/mochify.js/node_modules/puppeteer/.local-chromium/linux-869685` 
 - on `npm test`, there's a warning `Error: Could not find expected browser (chrome) locally. Run `npm install` to download the correct Chromium revision (869685).`

m90:
> Did puppeteer 9 build correctly?

Ah ok, I didn't get that you were referring to this branch, sorry. I was thinking the Puppeteer upgrade was a drive-by and not a necessity in the context of the PR. I was referring to the latest build on current `master`.

Anyways, the situation seems mysterious: when I checkout `2a4f0a8` from your branch locally I can't even `npm ci`, it just stalls. When I run `npm i` it installs, but it will do things I do not understand to `package-lock.json` (i.e. changes). In this case, however, the tests pass just fine in Node 16 as well. I would assume the Puppeteer update is therefore not strictly necessary and its side effect of updating the lockfile somehow fixed the situation. npm fun :dizzy: 

Do you think it would make sense to try again with updating watchify only and having tests pass in all Node versions?

As for updating Puppeteer it would make sense as well, but I would be slightly worried about this fix https://github.com/puppeteer/puppeteer/pull/7060 introducing some unwanted behavior around the topic of request interception.

albertyw:
I split the original changes in this PR into three PRs so they could be tested independently.  There's #227 which is master with `npm install` which made some small refactoring changes in package-lock.json. #226 and #228 are rebased on top of #227 and they update watchify and pupeteer, respectively.  All three PRs pass, including this one, which is a surprise to me since the last time I created the same PR, `npm ci` failed.

On the `npm ci`/`npm install` issues, my guess is that there was some `node_modules` or other cache that needed to be busted.  `npm ci` may have failed because of `If dependencies in the package lock do not match those in package.json, npm ci will exit with an error, instead of updating the package lock.` (https://docs.npmjs.com/cli/v7/commands/npm-ci)

Feel free to hold on #228, given possible issues with puppeteer compatibitliy.

m90:
Thank you for untangling this and getting npm to behave.

I just had a another look at the changes this would introduce https://github.com/browserify/watchify/compare/v3.11.1...v4.0.0 and it seems that watchify 4 also updates Browserify to v17, which currently does not behave nicely with some features in Mochify: https://github.com/mantoni/mochify.js/issues/224

I don't think it is a problem right now but I would still like to double check what happens here: https://github.com/mantoni/mochify.js/blob/1e7c98d72091a3d7cb69a47251627cc50a2ccfc2/lib/mochify.js#L264 and if this might somehow interfere with the v16 Mochify is using itself.

---

Strangely, looking at the code of watchify it seems, browserify is only depended upon for getting the same CLI interface here: https://github.com/browserify/watchify/blob/7b27a3c0d6bb796cef71faebfeb610ec62ad97b3/bin/args.js#L1 so I guess we should be fine here. I need to run now, but I hope we can get this merged soon. Thank you.

m90:
Ok, I double checked again and it seems watchify as a library never imports the browserify it requires so this will work just fine, with the only downside of installing Browserify twice when installing Mochify, which I guess is ok-ish.

Thank you!

m90:
:package: Released this as `8.0.1` :rocket:

PR #230, opened 2021-07-06
Add playwright driver

m90:
This is so eerily similar to `puppeteer` as of now, I wonder if it should share some code or if we're ok with the duplication.

Dunno about your plans, but having seen how easy this is, I would assume it makes more sense to wait with merging this while you still figure out the API and things like the testing story.

This currently works fine (running very basic stuff) with all 3 engines.

### DIFF
--- a/driver-playwright/index.html
+++ b/driver-playwright/index.html
@@ -1,0 +1,2 @@
+<!DOCTYPE html>
+<html><meta charset="utf-8"><body></body></html>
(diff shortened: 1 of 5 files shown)


mantoni:
Great! I was hoping to simplify this part of mochify, so that we can easily offer more options to run the tests.

I think the duplication is fine. There are some small edge cases already and it's good if we stay flexible with changes and updates and don't mix logic of the drivers.

mantoni (approved):
I guess this can be merged as is, right? We can keep iterating in separate PRs, if needed.

PR #232, opened 2021-07-10
Use non-blocking loop to poll for events

m90:
When debugging why the `jsdom` driver that I am working on was never running any tests I found out it's because of the following:

- in this setup `mochify` and `jsdom` run on the same thread
- the infinite loop that currently polls for events from the mochify reporter is blocking
- jsdom never gets around to execute any code because it's being blocked by the mentioned for loop

This is not a problem for headless browsers et al as they run on a different thread, yet I would assume allowing drivers to use the same thread as mochify can be helpful.

---

The [jsdom driver](https://github.com/mantoni/mochify.js/tree/jsdom-driver) works now including this change, but there's a lot of clean up to do still, so I thought this should go into a dedicated PR.

### DIFF
--- a/mochify/lib/poll-events.js
+++ b/mochify/lib/poll-events.js
@@ -3,22 +3,28 @@
 exports.pollEvents = pollEvents;
 
 async function pollEvents(driver, emit) {
-  for (;;) {
-    const events = await driver.evaluateReturn('mocha.mochify_pollEvents()');
-    if (!events) {
-      continue;
-    }
-    for (const [event, data] of events) {
-      if (event === 'mochify.callback') {
-        return data.code || 0; // stop polling
+  let interval;
+  const result = await new Promise((resolve) => {
+    interval = setInterval(async () => {
+      const events = await driver.evaluateReturn('mocha.mochify_pollEvents()');
+      if (!events) {
+        return;
       }
-      if (event === 'mochify.coverage') {
-        global.__coverage__ = data;
-      } else if (event.startsWith('console.')) {
-        console[event.substring(8)](...data);
-      } else {
-        emit(event, data);
+      for (const [event, data] of events) {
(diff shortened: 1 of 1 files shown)


mochify/lib/poll-events.js
```diff
@@ -3,22 +3,28 @@
...
+          global.__coverage__ = data;
+        } else if (event.startsWith('console.')) {
+          console[event.substring(8)](...data);
+        } else {
+          emit(event, data);
+        }
       }
-    }
-  }
+    }, 1);
```
mantoni, line 26: I guess this can lead to concurrent polling requests. Can you change it to poll, and then run the next poll slightly deferred? Maybe `setImmediate`?

m90: Yeah, sounds less risky to do it that way. Updated.

mochify/lib/poll-events.js
```diff
@@ -2,23 +2,32 @@
...
-    if (!events) {
-      continue;
-    }
-    for (const [event, data] of events) {
-      if (event === 'mochify.callback') {
-        return data.code || 0; // stop polling
+function pollEvents(driver, emit) {
+  return new Promise((resolve) => {
+    async function doPoll() {
+      const events = await driver.evaluateReturn('mocha.mochify_pollEvents()');
```
m90, line 8: One thing I do not understand in the current architecture yet: why does a driver need to expose a returning and a non-returning `evaluate`? Couldn't `evaluate` always return and then the caller can either discard or use the value?

mantoni: That is how I did it first, but I couldn't get it to work with driver-webdriver. It wraps the passed string in an anonymous function, and Puppeteer doesn't. If you find out how we can get away with a single function, I'm all for it.

mantoni (approved):
👍🙏

PR #233, opened 2021-07-11
Add jsdom driver

m90:
This adds a driver backed by `jsdom` which might be nice for environments that cannot run headless browsers or similar.

This seems to mostly work, apart from one issue where I'll comment inline.

### DIFF
--- a/driver-jsdom/index.js
+++ b/driver-jsdom/index.js
@@ -30,9 +30,14 @@ async function mochifyDriver(options = {}) {
   }
 
   virtual_console.on('jsdomError', (error) => {
+    process.exitCode = 1;
+    if (error && error.type === 'unhandled exception') {
+      // These errors will be logged by the global onerror handler
+      // in the client script, so there's no need to duplicate them here.
+      return;
+    }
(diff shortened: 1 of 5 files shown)


driver-jsdom/index.js
```diff
@@ -0,0 +1,47 @@
...
+    runScripts: 'dangerously',
+    url: options.origin,
+    virtualConsole: virtual_console
+  });
+
+  function end() {
+    return null;
+  }
+
+  virtual_console.on('jsdomError', (error) => {
```
m90, line 32: This is kind of tricky: `jsdomError` will be triggered on general evaluation or runtime errors (which is why we want this), but _also_ for uncaught exceptions in the code it is running, in which case we currently print it twice as it's being handled by the client's `window.onerror` handler as well as this handler here.

mantoni: Does it throw a different type of error or is there anything on the error object that we can use identify this case?

m90: They indeed have a `type` property we can switch on. Good idea.

mantoni (approved):

PR #234, opened 2021-07-14
Add driver options

mantoni:
I was thinking how we could specify driver options and pass them down. In this example, I prefixed the `engine` option (used by the playwright driver) with `--driver-`.

I can think of these naming conventions:

- `--driver-${option}`
- `--${driver}-${option}`
- `--driver-${driver}-${option}`

Or we re-introduce the [subarg](https://www.npmjs.com/package/subarg) module and allow `--driver-options [ --engine webkit ]`.

What do you think?

### DIFF
--- a/cli/index.js
+++ b/cli/index.js
@@ -11,6 +11,9 @@ const opts = yargs(hideBin(process.argv))
   .option('driver', {
     type: 'string'
   })
+  .option('driver-option', {
+    type: 'object'
+  })
   .option('reporter', {
     alias: 'R',
     type: 'string'
@@ -26,6 +29,7 @@ const opts = yargs(hideBin(process.argv))
 // The bundle command to run (from config):
 const config = {
   driver: opts.driver,
+  driver_options: opts['driver-option'],
   bundle: opts.bundle,
   reporter: opts.reporter,
   serve: opts.serve,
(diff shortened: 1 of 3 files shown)


m90:
Hmm, tough call. Using subarg would seem way more flexible at first glance (drivers could change their options and the CLI wouldn't have to be updated at all), but then this raises the question whether drivers would need to / should validate the given options and if we're ok with not having these documented in `--help` or similar?

mantoni:
Good call on `--help`, didn't think of that. I went with `yargs` because it's already a dependency of mocha. It allows us to generate the `--help` output the same way as mocha, but it wouldn't have support for `subarg`, or wouldn't know how to do it.

However, reading the `yargs` documentation, it has it's own concept of building objects. We could use something like `--driver-option.engine=webkit` which will parse to `{ driver-option: { engine: 'webkit' } }`.

See https://github.com/yargs/yargs/blob/master/docs/tricks.md#objects

m90:
> We could use something like --driver-option.engine=webkit which will parse to { driver-option: { engine: 'webkit' } }.

I think this sounds like a decent option if we get it for free with what we have already :+1:

mantoni:
`--decent-option` 😆

mantoni:
Changed to use the `yargs` dot notation to create the driver options.

Example:

```bash
cli/index.js test.js --driver playwright --driver-option.engine webkit
```

m90 (approved):
:sparkles: brilliant.

PR #235, opened 2021-07-14
Remove --server option in favor of --serve

mantoni:
Simply to have fewer options, I thought we could do just with `--serve $PATH` and remove `--server`.

`--server` is the same as `--serve .`


m90 (approved):
This is really good!

PR #236, opened 2021-07-16
Webdriver with Safari iOS Simulator

mantoni:
## WebDriver with local Safari

I experimented using `safaridriver` directly (which comes with Safari since v13). It worked right out of the box:

```bash
safaridriver --port 4444
```

And then configure webdriver capabilities accordingly (see documentation change).

## WebDriver with local iOS Simulator

This got me excited, so I tried running tests in iOS Simulator with this approach, which I coulnd't get to work at first. The `executeScript` calls always failed with a security exception, regardless of whether I served the local HTML file or used `--serve .`.

The solution was to turn off `https` and use a `http` server instead. Works like a dream. I have no idea why exactly this is, but francly, I also don't remember why an `https` was important at the time. Whatever I tried running through this setup worked.

@m90 Do you recall why we used an `https` server in Mochify so far?


m90:
> Do you recall why we used an https server in Mochify so far?

I have no idea tbh, from looking at the git history it seems to have been introduced like this and never was changed. My initial guess would have been that it allows a "secure context" (https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) which unlocks certain DOM APIs (`window.crypto.subtle` for example), but it seems all `localhost` domains are also considered secure even when served through `http`. The only situation this might be breaking would probably be:

- code uses APIs that rely on a secure context
- localhost is aliased to a non-localhost domain

but tbh this sounds _very_ far fetched, so I guess it's ok to just use `http`.

N.B.: Chrome seems to push "https first" https://blog.chromium.org/2021/03/a-safer-default-for-navigation-https.html with version 94, but from what I understand this shouldn't affect us here.

m90 (approved):
I can't test this because I'm on Linux, code LGTM though.

PR #237, opened 2021-07-16
Improve jsdom setup

mantoni:
- Inline html page
- Pass url through and default to http://localhost
- Pretend to be visual so that requestAnimationFrame works
- Turn of strict SSL checks

### DIFF
--- a/driver-jsdom/index.js
+++ b/driver-jsdom/index.js
@@ -1,29 +1,23 @@
 'use strict';
 
-const fs = require('fs/promises');
-const path = require('path');
 const jsdom = require('jsdom');
 
 exports.mochifyDriver = mochifyDriver;
 
-async function mochifyDriver(options = {}) {
-  options = Object.assign(
-    {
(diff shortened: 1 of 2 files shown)


driver-jsdom/index.js
```diff
@@ -1,29 +1,23 @@
...
-  const host_document = await fs.readFile(options.url);
   const virtual_console = new jsdom.VirtualConsole();
-
-  const { window } = new jsdom.JSDOM(host_document, {
-    runScripts: 'dangerously',
-    url: options.origin,
-    virtualConsole: virtual_console
-  });
+  const { window } = new jsdom.JSDOM(
+    '<!DOCTYPE html>\n<html><body></body></html>',
```
m90, line 12: This might be something to add at a later point, but maybe this could be configurable as a driver option somehow still?

m90 (approved):

driver-jsdom/index.js
mantoni: Absolutely. Good point.

PR #239, opened 2021-07-30
Load config with cosmiconfig

mantoni:
This change allows to configure mochify in `package.json` and various config file formats.

Users can specify a file with the `--config` option. If a config file is specified and a default configuration is also present, the configs are merged.

Command line options override the config file options. A custom config file overrides settings from the default config.

### DIFF
--- a/mochify/lib/load-config.js
+++ b/mochify/lib/load-config.js
@@ -9,23 +9,19 @@ async function loadConfig(options) {
 
   const default_config_promise = explorer.search();
 
-  const config_path = options.config;
-  if (config_path) {
-    const [default_config, specified_config] = await Promise.all([
-      default_config_promise,
-      explorer.load(config_path)
-    ]);
-
-    const config = Object.assign(specified_config.config, options);
-    if (default_config) {
-      return Object.assign(default_config.config, config);
-    }
-    return config;
+  if (options.config) {
+    const specified = await explorer.load(options.config);
+    const config = Object.assign(specified.config, options);
+    return mergeWithDefault(default_config_promise, config);
   }
 
+  return mergeWithDefault(default_config_promise, options);
+}
+
+async function mergeWithDefault(default_config_promise, config) {
   const default_config = await default_config_promise;
   if (default_config) {
-    return Object.assign(default_config.config, options);
+    return Object.assign(default_config.config, config);
   }
-  return options;
+  return config;
 }
--- a/mochify/lib/load-config.js
+++ b/mochify/lib/load-config.js
@@ -1,0 +1,31 @@
+'use strict';
+
+const { cosmiconfig } = require('cosmiconfig');
+
+exports.loadConfig = loadConfig;
+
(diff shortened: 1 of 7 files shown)


mochify/lib/load-config.js
```diff
@@ -0,0 +1,38 @@
...
+  const config_from_file = config_path
+    ? await explorer.load(resolveConfig(config_path))
+    : await explorer.search();
+
+  if (config_from_file) {
+    Object.assign(config, config_from_file.config, config);
+  }
+}
+
+function resolveConfig(config_path) {
```
m90, line 32: Is this function really needed? If I understand the cosmiconfig docs correctly calling load on an inexistent file will fail anyways, so Mochify wouldn't even need to try to work around this. Or is this doing something else that I don't yet understand?

mochify/lib/load-config.js
```diff
@@ -0,0 +1,38 @@
...
+    ]
+  });
+
+  const config_path = config.config;
+  const config_from_file = config_path
+    ? await explorer.load(resolveConfig(config_path))
+    : await explorer.search();
+
+  if (config_from_file) {
+    Object.assign(config, config_from_file.config, config);
```
m90, line 28: What's the reason for passing `config` twice? Would `Object.assign({}, config_from_file.config, config);` be easier to understand?

mochify/lib/load-config.js
```diff
@@ -0,0 +1,38 @@
...
+
+const { cosmiconfig } = require('cosmiconfig');
+
+exports.loadConfig = loadConfig;
+
+async function loadConfig(config) {
+  const explorer = cosmiconfig('mochify', {
+    searchPlaces: [
+      'package.json',
+      '.mochifyrc',
```
m90, line 11: Just curious: what' the expected format for a naked rc file? JSON? Or would YAMLwork too?

m90 (approved):
Looks good and the precedence pattern makes sense I guess.

mochify/lib/load-config.js
mantoni: I shamelessly copied the code from `lint-staged`. I removed this for now as I don't see that it's needed either.

mochify/lib/load-config.js
mantoni: `loadConfig` wasn't returning the config, it was "patching" the given object. I refactored this now.

mochify/lib/load-config.js
mantoni: The cosmiconfig docs say it can be YAML or JSON:

> YAML is a superset of JSON; which means YAML parsers can parse JSON; which is how extensionless files can be either YAML or JSON with only one parser.

mantoni:
Thanks for the review. I have refactored the whole thing to be more straight forward, use the cosmiconfig defaults and implement merging of the default configs, specified configs and command line options.

m90 (approved):

Issue #229, opened 2021-07-04
Mochify Rewrite


mantoni:
# Let's rewrite mochify 🚀

The `rewrite` branch is an attempt to rethink mochify from the ground up.

## Modular setup to only install what is needed

Mochify is currently rather bloated and loads all sorts of dependencies that most projects might never use. We can use npm workspaces to keep everything in one repository and publish smaller modules. I reserved the `@mochify` npm namespace.

The rewrite introduces "drivers" which are modules that bridge between the mochify CLI and runtime environments to run tests. Here are some ideas for drivers:

- `driver-puppeteer` (first implementation exists)
- `driver-selenium` to use the selenium protocol (ex `min-wd`)
- `driver-chromedriver` could be a bridge for https://www.npmjs.com/package/chromedriver

## Externalise the bundler

Allow to configure a `bundle` script to run using `execa` (https://www.npmjs.com/package/execa), similar to what `lint-staged` does. This eliminates the dependency on browserify and allows to use mochify with other bundlers 🙌. And it removes a **lot** of the current complexity. 👍 

Instead of bundling mocha with the test files, the pre-built `mocha/mocha.js` bundle is loaded, then a Mochify "client" is loaded, and then the actual script is loaded separately (to retain the correct line numbers for source-maps).

Only the actually passed files need to be bundled. `mocha/mocha.js` and the Mochify `client.js` files don't have any dependencies.

## Untangle reporter, console, runner and coverage output

Mochify currently does a lot of output parsing. The desired reporter is installed in the browser and the output has to be separated from `console.log` statements and output from `puppeteer` or `nyc`.

Instead, use a custom reporter in the browser and pass `ndjson` over a communication channel to feed events into a Mocha reporter on the CLI side. This removes the need for excessive parsing since event names can be used to communicate different kinds of output.

## Support config files

E.g. `cosmiconfig` https://www.npmjs.com/package/cosmiconfig

---

## Help wanted

The [rewrite branch](https://github.com/mantoni/mochify.js/tree/rewrite) is a baiscally a runnable exampe of the above ideas, with no unit tests, documentation and a rough implementation. Please feel free to contribute ideas, PRs, ...

@m90 @mroderick @albertyw @fatso83

m90:
Woohoo :tada: this seems to be an amazing start. I love how much simpler this is.

Some uninformed feedback after a first glance:

## Zero config

[...]

PR #241, opened 2021-08-01
Add initial unit and integration tests for CLI and mochify

mantoni:
This adds a few unit tests for the `mochify` module and some integration tests for the `mochify` and `cli` modules and runs them with GitHub actions.

Ref #229

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: ['12.x', '14.x', '16.x']
+        node-version: ['14.x', '16.x']
 
     steps:
       - name: Checkout
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,26 +3,41 @@ name: Build
 on:
   push:
     branches:
-      - master
+      - rewrite
   pull_request:
     branches:
-      - master
+      - rewrite
 
 jobs:
   build:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version:
-          - '12'
-          - '14'
-          - '16'
+        node-version: ['12.x', '14.x', '16.x']
+
     steps:
-     - uses: actions/checkout@v2
-     - uses: actions/setup-node@v2
-       with:
-         node-version: ${{ matrix.node-version }}
-     - name: Install dependencies
-       run: npm ci
-     - name: Run tests
-       run: npm t
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Cache modules
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
+      - name: Install
+        run: npm ci
+      - name: Lint
+        if: matrix.node-version == '16.x'
+        run: npm run lint
+      - name: Prettier
+        if: matrix.node-version == '16.x'
+        run: npm run prettier:check
+      - name: Test
+        run: npm test
+      - name: Integration
+        run: npm run test:integration
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,6 +29,9 @@ jobs:
           key: ${{ runner.os }}-node-${{ hashFiles('package-lock.json') }}
           restore-keys: |
             ${{ runner.os }}-node-
+      - name: npm7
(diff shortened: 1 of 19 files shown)


.github/workflows/test.yml
```diff
@@ -3,26 +3,44 @@ name: Build
...
+          key: ${{ runner.os }}-node-${{ hashFiles('package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
+      - name: npm7
+        if: matrix.node-version != '16.x'
+        run: npm i -g npm@7
+      - name: Install
+        run: npm ci
+      - name: Lint
+        if: matrix.node-version == '16.x'
```
m90, line 38: Is this just for saving some CPU cycles (running it once will be enough) or is there some other reason to it?

.github/workflows/test.yml
```diff
@@ -3,26 +3,44 @@ name: Build
...
-       run: npm ci
-     - name: Run tests
-       run: npm t
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Cache modules
```
m90, line 25: Does this have an effect when using `npm ci`? Will `~/.npm` contain a global cache that is used on `npm ci`?

mochify/test/files.integration.js
```diff
@@ -0,0 +1,23 @@
...
+
+describe('files', () => {
+  it('passes files to bundle command', async () => {
+    sinon.replace(process.stdout, 'write', sinon.fake());
+
+    await mochify({
+      driver: 'jsdom',
+      reporter: 'json',
+      files: [`${__dirname}/fixture/passes.js`],
+      bundle: 'cat'
```
m90, line 14: I assume this will make the tests fail on Windows, but I guess that's ok?

cli/test/playwright.integration.js
```diff
@@ -0,0 +1,54 @@
...
+
+describe('playwright', () => {
+  it('passes', async () => {
+    const result = await execa(
+      '../../index.js',
+      [
+        '--driver',
+        'playwright',
+        '--driver-option.engine',
+        'chromium',
```
m90, line 15: Did Firefox do strange things in CI or is this just for speed?

.github/workflows/test.yml
mantoni: Yep. Build speed as well. An alternative could be running lint and prettier in a separate job, but then it needs a to run npm install again.

.github/workflows/test.yml
mantoni: It just speeds up installs dramatically, if no changes are in package-lock.json.

mochify/test/files.integration.js
mantoni: Good point. Any idea for an alternative?

cli/test/playwright.integration.js
mantoni: There is no Firefox in the GitHub action container installed. However, chromium is. We'd need to figure out how to install Firefox, but then I thought it's not important for the purpose of this test.

.github/workflows/test.yml
m90: I wonder if this could turn into a red herring for people that don't know about this and have a build that fails on Node 16, but passes on Node 14, and who then go spelunking in changelogs and everything instead of having a look at the build output.

Probably ok to do it like this still.

cli/test/playwright.integration.js
m90: I was under the impression that Playwright installs all supported engines on `postinstall` so I am surprised this does not work. It does however place the executable outside of `node_modules` (as opposed to puppeteer), so maybe this needs some npm caching and/or $PATH adjustments in our context. I can look at that myself if you don't feel like doing it here.

See:
- https://playwright.dev/docs/ci#github-actions
- https://playwright.dev/docs/ci#caching-browsers

mochify/test/files.integration.js
m90: Hmm, `cat` seems to work in Powershell, but not in `cmd.exe`: https://stackoverflow.com/a/52388383/797194

Not entirely sure if we want to roll our own or even install a third-party package tbh.

cli/test/playwright.integration.js
mantoni: I might have changed too many things at once when I tried to get the build working.

mochify/test/files.integration.js
mantoni: Let's keep it like this until it's actually someone's problem.

cli/test/playwright.integration.js
mantoni: I gave this another try and integration tests time out when I configure firefox as the engine. I'll merge this now and we can look into it separately. Once we have a solution locally and for GitHub Actions, we can document the setup in `driver-playwright/README.md`.

PR #243, opened 2021-09-04
Allow to define "spec" in config and fix package.json

mantoni:
- The cli was missing the bin config.
- The playwright and puppeteer drivers where missing `index.html` in the `files` array.
- It wasn't possible to declare files to include in a config file. Mocha allows to define `spec`, which is what mochify supports now as well.

### DIFF
--- a/cli/index.js
+++ b/cli/index.js
@@ -1,8 +1,6 @@
 #!/usr/bin/env node
 'use strict';
 
-const { promisify } = require('util');
-const glob = promisify(require('glob'));
 const yargs = require('yargs/yargs');
 const { hideBin } = require('yargs/helpers');
 const { mochify } = require('@mochify/mochify');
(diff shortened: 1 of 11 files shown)


mantoni:
Sorry for not waiting for the review. I'm trying to use this in `@sinonjs/fake-timers` to fix a long pending issue with webdriver tests there, and I fear I won't have much time for this in the next couple of days.

m90:
> Sorry for not waiting for the review. 

Sorry for not catching things like a missing `bin` in the review I guess :facepalm: 

---

I just tried applying the new version to a project of mine and couldn't get it to work as the setup relies on a custom Chrome installation inside of a Docker container. I currently cannot pass any options to `driver-puppeteer` so I can't get it to work. Looking into how to fix it I wasn't entirely sure what's the best way forward:

- Should consumers be able to pass through all of puppeteer's `launchOptions`
- Should the driver selectively allow and pass certain options?

mantoni:
The `_retain` folder still has the old chromium implementation. You can check how it was done there. I noticed the same with GitHub actions that already have a chrome bin installed.

mantoni:
> Should consumers be able to pass through all of puppeteer's launchOptions

On the one hand, why not? On the other hand, there are quite a few unintuitive settings.

> Should the driver selectively allow and pass certain options?

We could provide sensible default settings for the testing scenario and then allow users to override those settings with any options they want.

PR #244, opened 2021-09-05
Rewrite: Fix webdriver for IE and Firefox

mantoni:
Converted the `client.js` to es5 to support IE.

Worked around an issue in the mozilla geckodriver: mozilla/geckodriver#1798


m90 (approved):
The underlying issue took me quite a while to understand. This seems to make sense.

PR #252, opened 2021-09-08
Allow passing through launch options to driver-puppeteer

### LINKED ISSUE #245: Rewrite: Puppeteer settings
Following up on #243, we need to allow custom chrome binaries and custom puppeteer settings.

We could provide sensible default settings for the testing scenario and then allow users to override those settings with any options they want.

For the custom chrome binary, this was the previous implementation:

```js
  if (opts.chrome) {
    options.executablePath = opts.chrome;
  } else if (process.env.PUPPETEER_EXECUTABLE_PATH) {
    // Workaround for https://github.com/puppeteer/puppeteer/issues/6957
    options.executablePath = process.env.PUPPETEER_EXECUTABLE_PATH;
  }
```

m90:
Resolves #245 

I opted for allowing users to pass through all of the available `launchOptions` to puppeteer while still applying some sane defaults. I have to admit I found writing the docs for this kind of hard, so maybe that's a sign this could be done better?

### DIFF
--- a/driver-puppeteer/index.js
+++ b/driver-puppeteer/index.js
@@ -7,19 +7,19 @@ exports.mochifyDriver = mochifyDriver;
 const default_url = `file:${__dirname}/index.html`;
 
 async function mochifyDriver(options = {}) {
-  const stderr = options.stderr || process.stderr;
+  const { stderr = process.stderr, ...launch_options } = options;
+
+  launch_options.args = [
+    '--allow-insecure-localhost',
+    '--disable-dev-shm-usage',
+    ...(launch_options.args || [])
+  ];
 
   const browser = await driver.launch({
-    ignoreHTTPSErrors: true
-    /*
-    args: [
-      '--allow-insecure-localhost',
-      '--disable-dev-shm-usage',
-      '--no-sandbox',
-      '--disable-setuid-sandbox',
-      '--disable-web-security'
-    ]
-    */
+    ignoreHTTPSErrors: true,
+    // Workaround for https://github.com/puppeteer/puppeteer/issues/6957
+    executablePath: process.env.PUPPETEER_EXECUTABLE_PATH,
+    ...launch_options
   });
 
   const page = await browser.newPage();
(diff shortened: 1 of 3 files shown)


driver-puppeteer/index.js
```diff
@@ -7,19 +7,19 @@ exports.mochifyDriver = mochifyDriver;
 const default_url = `file:${__dirname}/index.html`;
 
 async function mochifyDriver(options = {}) {
-  const stderr = options.stderr || process.stderr;
+  const { stderr = process.stderr, ...launch_options } = options;
+
+  launch_options.args = [
+    '--allow-insecure-localhost',
```
m90, line 13: This would be the default behavior of Mochify 8.

driver-puppeteer/package.json
```diff
@@ -23,6 +23,9 @@
   "peerDependencies": {
     "@mochify/mochify": "^0.1.0"
   },
+  "engines": {
+    "node": ">=12.0.0"
```
m90, line 27: This is needed so that eslint does not choke on destructuring.

driver-puppeteer/index.js
```diff
@@ -7,19 +7,19 @@ exports.mochifyDriver = mochifyDriver;
...
-      '--disable-dev-shm-usage',
-      '--no-sandbox',
-      '--disable-setuid-sandbox',
-      '--disable-web-security'
-    ]
-    */
+    ignoreHTTPSErrors: true,
+    // Workaround for https://github.com/puppeteer/puppeteer/issues/6957
+    executablePath: process.env.PUPPETEER_EXECUTABLE_PATH,
+    ...launch_options
```
mantoni, line 22: Does this allow to override `executablePath` via the `launch_options`?

m90: Yup, that would still work:

```console
➜  ~ FOO_BAR=some_value node
Welcome to Node.js v14.15.5.
Type ".help" for more information.
> const opts = { foo_bar: 'other_value' }
undefined
> { foo_bar: process.env.FOO_BAR, ...opts }
{ foo_bar: 'other_value' }
```

We can also use `Object.assign` though if that's easier to understand?

mantoni (approved):

mantoni:
Released 🎉

PR #253, opened 2021-09-08
Allow external drivers

### LINKED ISSUE #248: Rewrite: Allow "external" mochify drivers
Currently only the mochify drivers in this project can be used due to the way they are loaded.

We could allow using "external" drivers by try-catching module load errors in the require statement with our naming convention, and fall back to the provided driver name as is.

m90:
Closes #248


mantoni (approved):
Nice 👍

PR #254, opened 2021-09-08
Cater for cases when puppeteer driver args is a string only

m90:
This is a fixup for #252

It seems that when passing multiple values through the CLI like:

```
mochify --driver-options.args='--allow-chrome-as-root' --driver-options.args='--no-sandbox'
```

`yargs` will pass  `['-allow-chrome-as-root',  '--no-sandbox']` when passing a single value like

```
mochify --driver-options.args='--allow-chrome-as-root'
```

will yield the plain string `'--allow-chrome-as-root'`

### DIFF
--- a/driver-puppeteer/index.js
+++ b/driver-puppeteer/index.js
@@ -12,10 +12,7 @@ async function mochifyDriver(options = {}) {
   // In case this arrives through CLI flags, yargs will pass a string
   // when a single arg is given and an Array of strings when multiple
   // args are given.
-  const extra_args = Array.isArray(launch_options.args)
-    ? launch_options.args
-    : [launch_options.args].filter(Boolean);
-
+  const extra_args = [].concat(launch_options.args).filter(Boolean);
   launch_options.args = [
     '--allow-insecure-localhost',
     '--disable-dev-shm-usage',
--- a/driver-puppeteer/index.js
+++ b/driver-puppeteer/index.js
@@ -9,10 +9,17 @@ const default_url = `file:${__dirname}/index.html`;
 async function mochifyDriver(options = {}) {
   const { stderr = process.stderr, ...launch_options } = options;
 
+  // In case this arrives through CLI flags, yargs will pass a string
+  // when a single arg is given and an Array of strings when multiple
+  // args are given.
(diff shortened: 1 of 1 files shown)


driver-puppeteer/index.js
```diff
@@ -9,10 +9,17 @@ const default_url = `file:${__dirname}/index.html`;
 async function mochifyDriver(options = {}) {
   const { stderr = process.stderr, ...launch_options } = options;
 
+  // In case this arrives through CLI flags, yargs will pass a string
+  // when a single arg is given and an Array of strings when multiple
+  // args are given.
+  const extra_args = Array.isArray(launch_options.args)
+    ? launch_options.args
+    : [launch_options.args].filter(Boolean);
```
m90, line 17: The filter is here so we get `[]` in case `args` is not defined. If this is all too smart, lmk.

mantoni: How about `[].concat(args)`?

m90: Oh yeah right, I always forget about the funny animal that is `concat`. Let me change that real quick.

m90: Hmm, this would still give us `[undefined]` in case no `args` are given?

m90:
Good news is that including this fix I can run all of my existing Mochify 8 suites just fine using `@mochify/cli` and the puppeteer driver, both in Docker and unboxed on my local setup.

driver-puppeteer/index.js
mantoni: Ah, yes. So `args ? [].concat(args) : []` maybe?

mantoni: Or how it's done in browserify (where I learned this trick):

```js
[].concat(args).filter(Boolean)
```

m90: Done so.

mantoni (approved):
👍

mantoni:
📦 Released in `@mochify/driver-puppeteer` `v0.2.2`.

m90:
Nice, this works well for me now when porting an existing mochify setup. Local, local in docker and in CircleCi all run well.

PR #256, opened 2021-09-09
Deep merge config objects

### LINKED ISSUE #255: Rewrite: nested configuration does not get merged correctly
When defining driver options we might run into situation where you have a command that relies on runtime information like:

```
mochify src/{,**}/*.test.js --config .mochifyrc.json --driver-option.executablePath $(which google-chrome)
```

And a config file like:

```
{
  "bundle": "browserify",
  "driver": "puppeteer",
  "driver_options": {
    "args": ["--no-sandbox", "--allow-chrome-as-root"]
  }
}
```

Right now, this does not seemt o work correctly as the cli args will override the options given in the config file entirely instead of merging them.

m90:
Fixes #255 

Introducing the `deepmerge` dependency is kind of the nuclear option here, but I found myself hunting way too many edge cases when trying to implement this myself. If you prefer to solve this without a dependency let me know and I can see what I can get figured out.

### DIFF
--- a/mochify/lib/load-config.js
+++ b/mochify/lib/load-config.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const deepmerge = require('deepmerge');
 const { cosmiconfig } = require('cosmiconfig');
 
 exports.loadConfig = loadConfig;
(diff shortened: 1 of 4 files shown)


mantoni (approved):
The extra dependency is fine with me. Thank you for looking into this! 👍

mantoni:
📦 Released in `@mochify/mochify` `v0.2.1`.

PR #257, opened 2021-09-10
Import fs/promises in a way that is compatible with Node 12 still

m90:
From what I understand we still support Node 12 where importing `fs/promises` fails and needs to be `require('fs').promises` instead. See: https://nodejs.org/api/fs.html#fs_promises_api

image

Alternatively we could of course also drop support for 12 for the rewrite, but then again I really hope we can get this shipped before its EOL (which is end of April 2022).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: ['14.x', '16.x']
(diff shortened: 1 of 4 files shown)


mantoni:
Can you add node 12.x to the build matrix?

m90:
> Can you add node 12.x to the build matrix?

Done so, passes as well.

mantoni (approved):
👍

PR #260, opened 2021-09-10
Shutdown on errors

mantoni:
This is an alternative implementation of #259 which retains parallel execution of the driver and the bundler. These are the two sub processes that may take a while, so I would prefer to not run them sequentially.

One of the integration tests highlighted that the `driver-jsdom` doesn't return a promise, which this fixes as well.

### DIFF
--- a/driver-jsdom/index.js
+++ b/driver-jsdom/index.js
@@ -38,9 +38,9 @@ function mochifyDriver(options = {}) {
     return Promise.resolve(window.eval(script));
   }
 
-  return {
+  return Promise.resolve({
     evaluate,
     evaluateReturn: evaluate,
     end
-  };
+  });
 }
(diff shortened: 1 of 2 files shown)


m90:
Ah yeah, I just added a comment about how I tried this approach before but didn't like it :)

You're totally right about how both bundling and launching browser automation can be a time consuming task so let's do it like this even if it makes the code a little more complicated.

driver-jsdom/index.js
```diff
@@ -38,9 +38,9 @@ function mochifyDriver(options = {}) {
     return Promise.resolve(window.eval(script));
   }
 
-  return {
+  return Promise.resolve({
```
m90, line 41: Good :fishing_pole_and_fish:

m90 (approved):

mantoni:
I made new patch releases for `@mochify/mochify` and `@mochify/driver-jsdom`.

PR #261, opened 2021-09-10
Return exit code from main API method instead of setting it

m90:
This picks up https://github.com/mantoni/mochify.js/pull/259#issuecomment-916875460

Technically this is a breaking change, so maybe you want to wait with merging this before we do a `0.3.0`?

### DIFF
--- a/mochify/index.js
+++ b/mochify/index.js
@@ -59,9 +59,7 @@ async function mochify(options = {}) {
     await shutdown(driver, server);
   }
 
-  if (exit_code) {
-    process.exitCode = exit_code;
-  }
+  return { exit_code };
 }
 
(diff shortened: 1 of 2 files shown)


mochify/index.js
```diff
@@ -59,9 +59,7 @@ async function mochify(options = {}) {
     await shutdown(driver, server);
   }
 
-  if (exit_code) {
-    process.exitCode = exit_code;
-  }
+  return { exit_code };
```
m90, line 62: If my understanding of a catch-less `try / finally` is correct we do not need to set a default value here as we never get here in case `run` throws, right?

mantoni (approved):
Technically yes. But I guess at this early stage we can also just make patches. Nobody will notice 🤫

Issue #249, opened 2021-09-06
Rewrite: Support native JavaScript modules


mantoni:
We have the `--serve <dir>` option already, which loads the given web page in the browser before injecting tests.

We could add a `--inject-modules` option (any better names?) that adds a `<script type="module">` tag in the generated start page for each given file.

m90:
I think I would like to look into this (in case noone else started yet) as, quite frankly, I know nothing about ESM at all and it starts to get embarassing. Maybe I can use this to get up to speed.

There's one thing I do not fully understand here yet (probably a good demonstration of how I know nothing about ESM):

Assuming someone runs the following command:

```
mochify ./src/foo.test.js --driver puppeteer --inject-modules
```

and `src/foo.test.js` looks like this:

```js
import * as assert from 'assert'
import * as foo from './foo.js'

describe('foo', function () {
  it('does not throw when calling bar', function () {
    assert.doesNotThrow(function () {
      foo.bar()
    })
  })
})
```

Mochify would then inject (I think?):

```html
<script>/* mocha ... */</script>
<script>/* client ... */</script>
<script type="module">
  /* above test case */
</script>
```

What I am failing to understand now is the following: how does the test case resolve:
1. modules like `assert` or other libraries without bundling
2. source files like `./foo.js`?

Is 1 even possible? Would 2 work out of the box in case users would pass `--serve ./src` or would this require extra setup to be taken?

mantoni:
> modules like assert or other libraries without bundling

That wouldn't work. There are already other solutions that generate mini-bundles on request. We could pass requested files through a bundler, if one is specified. Not sure here and it might be out of scope for Mochify.

> Would the source document instead need to load the files something like <script src="/foo.test.js" type="module"></script>?

This is how I would do it, yes. A server is required anyway to be able to import anything with esm. I guess generating the script tags is all that is needed for now.

PR #264, opened 2021-09-11
Allow loading tests as esm modules

m90:
This is a first attempt at https://github.com/mantoni/mochify.js/issues/249

This will currently just time out when using the JSDOM driver, not entirely sure if it's not supported or if it's my fault yet (leaving it a draft until then, but feel free to feedback already if you have some). I would also like to add some unit tests to this, especially around transposing file paths against the server root.


mochify/client.js
```diff
@@ -133,17 +133,25 @@
...
+      document.body.appendChild(s);
+    }
     // Run mocha
-    mocha.run(function (code) {
-      if (typeof __coverage__ !== 'undefined') {
-        write('mochify.coverage', window.__coverage__);
-      }
-      write('mochify.callback', { code: code });
-    });
+    Promise.all(window.mochify_pending || [])
```
m90, line 143: Do all browsers we need to support (you downgraded the client file recently) support native Promises?

cli/index.js
```diff
@@ -28,6 +28,9 @@ const opts = yargs(hideBin(process.argv))
   .option('server-option', {
     type: 'object'
   })
+  .option('esm', {
```
m90, line 31: Is this a good name?

m90:
Ok, it seems this is not supported by JSDOM https://github.com/jsdom/jsdom/issues/2475 which is probably good an bad at the same time.

This opens up an interesting question though: in case we would like to have a driver throw in such a case (something is not supported by that particular driver), how does the driver even know about it? Right now, all it ever sees is the `driverOptions`. Does this mean we also need to pass the `options` themselves to driver so they can validate them? Or do we rely on documenting things like this?

mochify/client.js
mantoni: I guess they don't. I wanted to get things running down to IE, since SinonJS still has that supported.

We might say we don't want to go that far back and allow ourselves to feel like it's 2021 😁

Anyway, I think there's a misunderstanding in where this is implemented. I thought the script tags would be generated in start page of the http server. Do you think it's better to do this in the client?

m90: Yeah I initially understood it just like you described too, but then I found it meant we'd have to reimplement it over and over at driver level (in case I do understand it correctly that is 🙉 ). Like this it could happen at Mochify level once.

Then again, doing it at driver level of course would also solve the problem of knowing if the driver can actually support it.

---

That being said, I think we'd need that promise like thing even in case we implemented this feature at driver level as we'd still need to make sure script tags have loaded all of their dependencies before we run Mocha. Or is that incorrect?

mochify/lib/resolve-modules.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+const path = require('path');
+
+exports.resolveModules = resolveModules;
+
+function resolveModules(serve, files) {
```
mantoni, line 7: Is this necessary? Couldn't the user simply launch the server in the directory that hosts the files?

mantoni (commented):
Wow. That's a lot more code than I expected. What where the issues with simply generating script tags in the server landing page for the `files`? I naively thought that's all that is necessary 🤔

mochify/client.js
m90: Or would using `esm` mean we generate a templated index document in the API module that is served by the server and then used as the host for running the test cases?

mantoni:
> Does this mean we also need to pass the options themselves to driver so they can validate them? Or do we rely on documenting things like this?

I didn't think about this case yet. Good question. Up for discussion, but I'd wait until we have at least two of those cases. That makes it easier to see a good solution.

mochify/client.js
mantoni: Yeah, I though we generate the script tags in the server in `@mochify/mochify`. Driver independent and without changing the client.

m90: It sounds good although I am not entirely sure we can get away without having to change the client. Let me try doing that.

mochify/lib/resolve-modules.js
m90: From what I understand it is necessary. E.g. someone runs 

```
mochify ./src/thing.test.js --esm --serve ./src --driver puppeteer
```

which means that without comparing the paths, we'd inject `<script src="./src/thing/test.js"></script>` which would 404 as the server root is already serving from `./src`. It would of course work if we'd just serve from `process.cwd()` in case `--esm` is passed, but then again that could cause a lot of confusion. Not sure, I think I would err on the side of doing the path munging just to be sure.

m90: Maybe we could also just disregard `--serve` when `--esm` is being passed, assuming `process.cwd()` as the root and moving on. Would also make it as easy as

```
mochify ./src/thing.test.js --esm --driver puppeteer
```

mantoni: I'm not sure either. Let's get it work first and then decide on how we resolve the paths. It's just a detail.

mochify/client.js
m90: Turns out you were right (I don't understand why module loading happens asynchronously in case you append a script tag though) and we don't need to touch the client when we do it at server level. Feel free to have another look and let me know what you think.

mochify/index.js
```diff
@@ -18,26 +18,40 @@ async function mochify(options = {}) {
   const mocha_runner = createMochaRunner(config.reporter || 'spec');
   const { mochifyDriver } = resolveMochifyDriver(config.driver);
 
+  const [mocha, client, files] = await Promise.all([
+    readFile(require.resolve('mocha/mocha.js'), 'utf8'),
+    readFile(require.resolve('./client'), 'utf8'),
+    resolveSpec(config.spec)
+  ]);
+
+  const configured_client = setupClient(client, config);
```
m90, line 27: It seems we previously never passed `config` to that call so I added it as a drive by.

mochify/index.js
```diff
@@ -18,26 +18,40 @@ async function mochify(options = {}) {
...
+
   const driver_options = config.driver_options || {};
   let server = null;
-  if (config.serve) {
-    server = await startServer(config.serve, config.server_options);
+  if (config.serve || config.esm) {
+    const scripts = [];
+    if (config.esm) {
+      scripts.push({ content: mocha });
```
m90, line 34: I wonder if we should always serve them directly, even when users are not using the `esm` option?

mochify/lib/resolve-modules.js
m90: I left it out for now. If we want it back it should be easy to bring back.

mochify/index.js
mantoni: That's a very good point. Would make the file concatenation go away if no bundler is specified, and give correct stack traces without a source map as well.

mantoni: Although this means there is always a server, no?

m90: I think we have 3 options:
- include the "prelude" in the served doc when `--esm` is given (as this requires a server)
- include it in the served doc when `--esm` or `--server` is given (as both require a server)
- always run a server, always include it in the doc

I have to admit I am not entirely sure about all of the consequences choosing option 3 would have so right now my answer would be "i don't know", but I will think about it more.

mochify/lib/server.js
```diff
@@ -25,11 +25,25 @@ async function startServer(base_path, options = {}) {
...
   return async (req, res) => {
     if (req.url === '/') {
       res.writeHead(200);
-      res.end('<!DOCTYPE html>\n<html><body></body></html>');
+      res.end(`
+<!DOCTYPE html>
+<html>
+${scripts
+  .map((script) => {
+    const { content = '', ...attrs } = script;
```
mantoni, line 37: Why did you come up with this "protocol"? Couldn't we add the generated script tag strings to the `scripts` array?

mochify/index.js
mantoni: I just realized that option 3 is not a good idea. Drivers that inject scripts into remote computers wouldn't work (e.g. `webdriver`).

mochify/lib/server.js
m90: > Why did you come up with this "protocol"? 

It's called overengineering and I am guilty :church: 

> Couldn't we add the generated script tag strings to the scripts array?

Then who is responsible for distinguishing a string like `./foo.test.js` and `alert('hello world')`? Even if we'd have `index.js` pass down strings the server could use as-is, someone would still need to know about the logic how to construct valid inline scripts and scripts that reference source files.

---

We could definitely make this a little less complicated though I guess. What do you think about passing `scripts` (for inline scripts) and `modules` (for source files)? This way we could still have the "template" figure out how to render these things (which is where I think this should happen).

m90: I changed it to work as described and I think it's a lot nicer now.

mochify/index.js
m90: I changed it so that we use option 2 which I guess makes the most sense.

mantoni (approved):
This looks pretty good to me. 🙌

PR #266, opened 2021-09-12
Rewrite: 11ty website

### LINKED ISSUE #265: Rewrite: Documentation
If we want to flip the switch and bring the rewrite to a 1.0.0 we'll also need documentation. I'd be happy to contribute here as well, so I'll just ask some questions:

## READMEs, Wiki, Docs site?

I would assume the Rewrite requires some more detailed docs than the original Mochify. How far do we want to go here? Do we put everything into READMEs in the repository, do we start using the repo's wiki or do we even go for a simple docs site? I have to admit I quite enjoy reading non-README documentation by now, but I also see the maintenance burden this brings. Wikis are odd, but that's very subjective.

## All-in-one or per module

Is there a single entrypoint for all `@mochify` documentation or does each package get its own docs? I think having everything in one place would be nice, but this of course creates the problem of not having docs (unless you start scripting) on npmjs.org (although I have to admit that I personally will never read these there but always click through to the backing repository)

### Scope of driver docs

What's the scope of driver specific documentation? Does this really only cover the driver itself or would it also tell you about how to use the driver in conjunction with Mochify itself?

### CLI docs

Do CLI docs live in the top level README or do they go into the `cli` package?

### Top level README

What should someone who has just read the top level README be able to take away? Get Mochify set up somehow or find answers to detailed questions as well?

## CLI help

Right now `--help` does not really yield anything helpful although it should:

```
Options:
      --help           Show help                                       [boolean]
      --version        Show version number                             [boolean]
      --config                                                          [string]
      --driver                                                          [string]
      --driver-option
  -R, --reporter                                                        [string]
      --bundle                                                          [string]
      --esm                                                            [boolean]
      --serve                                                           [string]
```

How "fancy" should this get? Do we want to include usage examples in the output (which I like a lot personally)?

mantoni:
This is a suggestion how we could combine per module `README` files with a "All-in-one" documentation page. The proposal uses `eleventy` to generate a static website using the `README` files as content.

For this to work, we need to add frontmatter to the README files. If this is something we want to avoid, we could alternatively store the information in a file called `README.11tydata.json` next to each `README`.

We could create a GitHub action to deploy this on GitHub pages, following the steps described here: https://www.linkedin.com/pulse/eleventy-github-pages-lea-tortay/

## Run locally

1. Check out this branch
2. `npm ci`
3. `npm start`
4. Open http://localhost:8080 in your browser.

Closes #265


m90:
I think the overall structure makes a lot of sense like this.

> For this to work, we need to add frontmatter to the README files. If this is something we want to avoid, we could alternatively store the information in a file called README.11tydata.json next to each README.

I think that might be worth it, even if it's only for the aesthetics.

mantoni:
A different approach could be to replicate the structure in `site` and keep the website entirely separate from the modules. This way the readme files can have different content than the website.

That would also give us more flexibility regarding the website design. What do you think?

m90:
> This way the readme files can have different content than the website.

I'm not entirely sure I understand this. Would this mean we would have a README in the module root and a totally different file in `site`? Or are they somehow (sym)linked, derived, processed or similar?

mantoni:
Currently I'm using the readme files to generate the content of the website. We could, alternatively, generate the content from a set of entirely different markdown and template files. They don't have to be the same. The website could live in `site` and not use any content from the module folders.

My question is: Do we want the readme files to be different, maybe shorter, or do we want the same content anyway?

m90:
> We could, alternatively, generate the content from a set of entirely different markdown and template files. 

How about we start by having everything in `site` but symlinking the existing READMEs into this directory up until the point we find that we need different content? If I understand it correctly this would also allow us to have frontmatter-less READMEs and define the `foo.11tydata.json` files in `site`?

mantoni:
Interesting idea to symlink the readme files. I'll give that a try. 👍

mantoni:
The symlinks with `index.11tydata.json` files work fine. Thank you for the suggestion!

PR #267, opened 2021-09-15
Rewrite: Add help and example content for CLI module

m90:
This is a first pass at adding some content and structure to the output of `mochify --help`. Right now, this is relatively wordy (maybe a little too wordy) as it's the only documentation around, I think it should be possible to reduce the amount of inline docs over time (moving it to `@mochify/mochify`).

### DIFF
--- a/cli/index.js
+++ b/cli/index.js
@@ -9,8 +9,8 @@ const opts = yargs(hideBin(process.argv))
   .usage(
     '$0 [options] <spec...>',
     'Run Mocha tests in real browsers.',
-    (local) => {
-      local
+    (cmd) => {
+      cmd
         .example(
           '$0 --driver puppeteer --bundle browserify "./src/**/*.test.js" ',
           'Bundle all files matching the given spec using browserify and run them using @mochify/driver-puppeteer.'
@@ -21,11 +21,11 @@ const opts = yargs(hideBin(process.argv))
         )
         .example(
           '$0 "./src/**/*.test.js" ',
-          'Run all tests matching the given spec using the configuration from package.json'
+          'Run all tests matching the given spec using the default configuration lookup.'
         )
         .example(
-          '$0 --config .mochifyrc.yml "./src/**/*.test.js" ',
-          'Run all tests matching the given spec using the configuration from .mochifyrc.yml'
+          '$0 --config mochify.webdriver.js "./src/**/*.test.js" ',
+          'Run all tests matching the given spec using the configuration from mochify.webdriver.js.'
         )
         .epilogue(
           `Mochify Resources:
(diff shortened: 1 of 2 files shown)


cli/index.js
```diff
@@ -6,31 +6,82 @@ const { hideBin } = require('yargs/helpers');
...
+    describe:
+      'Free form options to pass to the driver in use. Pass an arbitrary number of options using `--driver-option.foo 1 --driver-option.bar 2`. Refer to the documentation of the driver in use for available options.'
   })
   .option('reporter', {
     alias: 'R',
-    type: 'string'
+    type: 'string',
+    group: 'Options:',
+    describe:
+      'The Mocha reporter to use. Right now, only reporters that are included with Mocha itself can be used.'
```
m90, line 53: This might not be correct.

cli/index.js
```diff
@@ -6,31 +6,82 @@ const { hideBin } = require('yargs/helpers');
...
+        );
+    }
+  )
   .option('config', {
-    type: 'string'
+    alias: 'C',
+    type: 'string',
+    group: 'Options:',
+    describe:
+      'The config file to use. In case `package.json` is given, the configuration is expected to be stored in a top-level "mochify" key. In case an option is present in both the config file and as a CLI flag, the flag takes precedence. Refer to the documentation of `@mochify/mochify` for available configuration options.'
```
m90, line 33: Any opinions on handling long strings like this in regards to line length?

cli/index.js
```diff
@@ -6,31 +6,82 @@ const { hideBin } = require('yargs/helpers');
...
+    'Run Mocha tests using Mochify.',
+    (local) => {
+      local
+        .example(
+          '$0 --driver puppeteer --bundle browserify "./src/**/*.test.js" ',
+          'Bundle all files matching the given spec using browserify and run them using @mochify/driver-puppeteer.'
+        )
+        .example(
+          '$0 --config package.json "./src/**/*.test.js" ',
+          'Run all tests matching the given spec using the configuration from package.json\'s "mochify" key.'
```
mantoni, line 20: The `mochify` key in `package.json` is picked up automatically. Specifying it like this would be superfluous.

cli/index.js
```diff
@@ -6,31 +6,82 @@ const { hideBin } = require('yargs/helpers');
 const { mochify } = require('@mochify/mochify');
 
 const opts = yargs(hideBin(process.argv))
+  .usage(
+    '$0 [options] <spec...>',
+    'Run Mocha tests using Mochify.',
```
mantoni, line 11: "Run Mocha tests in real browsers."?

cli/index.js
mantoni: What happens if you use a multiline string? Does it screw up the layout?

Otherwise we can just concatenate strings with `+`.

cli/index.js
mantoni: True. This might actually work.

mantoni (commented):
Yeah, I agree that this is probably too wordy. Comparing it to the `mocha -h` output, they only have very short descriptions. I like the content you've written though. We could put it in the readme (and therefore on the website).

What I quite like about the mocha help output is that they link to the repo and the documentation website.

cli/index.js
m90: Oh that's interesting to know.

cli/index.js
```diff
@@ -8,80 +8,82 @@ const { mochify } = require('@mochify/mochify');
...
-          '$0 --config package.json "./src/**/*.test.js" ',
-          'Run all tests matching the given spec using the configuration from package.json\'s "mochify" key.'
+          '$0 --esm --reporter dot --driver puppeteer "./src/**/*.test.js" ',
+          'Run all tests matching the given spec as ES modules in puppeteer and use the "dot" reporter for output.'
         )
         .example(
-          '$0 --esm --reporter dot "./src/**/*.test.js" ',
-          'Run all tests matching the given spec as ES modules and use the "dot" reporter for output.'
+          '$0 "./src/**/*.test.js" ',
+          'Run all tests matching the given spec using the configuration from package.json'
```
mantoni, line 24: If there is a `mochify` key in `package.json`. cosmiconfig tries various options in a specific order. Maybe we should just document that with a link to cosmiconfig?

See https://github.com/davidtheclark/cosmiconfig#explorersearch

cli/index.js
```diff
@@ -8,80 +8,82 @@ const { mochify } = require('@mochify/mochify');
...
         );
     }
   )
   .option('config', {
     alias: 'C',
     type: 'string',
     group: 'Options:',
-    describe:
-      'The config file to use. In case `package.json` is given, the configuration is expected to be stored in a top-level "mochify" key. In case an option is present in both the config file and as a CLI flag, the flag takes precedence. Refer to the documentation of `@mochify/mochify` for available configuration options.'
+    describe: 'The config file to use (defaults to "package.json")'
```
mantoni, line 40: See above comment. There's no default.

cli/index.js
m90: Ah ok, that's not entirely obvious tbh (I remember seeing a list of filenames being removed in a commit once and thought this meant there is no default lookup happening, but apparently that's not the case). Will adjust this accordingly.

mantoni: Yes, I removed the file list because I discovered that I replicated the cosmiconfig defaults 😆

cli/index.js
```diff
@@ -6,31 +6,83 @@ const { hideBin } = require('yargs/helpers');
...
   .option('server-option', {
-    type: 'object'
+    type: 'object',
+    group: 'Options:',
+    describe: 'Pass options to the server (requires --serve or --esm)'
   })
+  .updateStrings({
+    'Options:': 'Other:'
+  })
+  .wrap(process.stdout.columns ? Math.min(process.stdout.columns, 80) : 80)
```
m90, line 85: This could also use higher threshold than '80` in case `process.stdout.columns` is given? Probably _very_ subjective.

mantoni: Behaving the same as mocha is generally a good approach.

mantoni (approved):
👍😎

Issue #268, opened 2021-09-16
Rewrite: read spec/bundle from stdin


m90:
We talked about this in #229 already, but tbh I am not entirely sure if it reads like a yes or a no over there, which is why I am opening this issue.

---

Proposal: Instead of passing a list of files and a bundling command to Mochify, the (possibly) bundled spec could also be read from `stdin`. This could work like:

## CLI

Users could invoke the CLI and pass `-` as the spec:

```
./my_custom_bundler_very_good_never_fails.sh | mochify --driver puppeteer -
```

which would just pipe the received bundle through to the driver.

## API

The API could be extended to either receive a string (files) or a stream (stdin) as the `spec` option:

```js
mochify({
  spec: process.stdin
})
```

This might also be helpful in gulp-like scenarios (if they still exist).

## How does this work in conjunction with `--bundle`?

The interesting question is how does this work in case `--bundle` is given. Does Mochify a. disregard the `--bundle` option (maybe it refuses to do things in the first place) or b. pipe the stream through to the bundle command itself?

## Do we even want this?

While I personally would prefer this for passing bundles (as a consumer I do not need to know about Mochify internals but can run my bundling command in total isolation), it also adds some extra complexity to Mochify. Use cases that use code generation or similar (if such even exist) could also just use temp files or similar instead of piping output.

mantoni:
I wouldn't use it this way, but I support this feature. How would this be enabled? How does the `cli` decide whether it reads from `stdin` or uses the files (files should actually default to `test` like mocha does it)?


Regarding `--bundle` I would probably throw an exception.

m90:
I think the convention is to just pass `-` (as in above example).

PR #269, opened 2021-09-16
Allow reading bundle from stdin

m90:
Solves #268

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -64,3 +64,21 @@ Files in the given directory will be served as static assets.
 
 Options to pass to the server in case `--serve` or `--esm` is being used.
 Currently only `--server-option.port` for passing the port to use is supported.
+
+## Spec
+
+The `spec` argument can be a list of files or a glob pattern that will be resolved by Mochify.
+
+```
+mochify ./src/foo.test.js ./src/bar.test.js
+mochify ./src/*.test.js # Let the shell handle glob expansion
+mochify "./src/*.test.js" # Let Mochify handle glob expansion
+```
+
+### Reading a bundle from `stdin`
+
+When given `-` as the spec, Mochify expects to read a bundled test suite from `stdin`:
+
+```
+browserify -t babelify ./src/*.test.js | mochify -
+```
--- a/cli/index.js
+++ b/cli/index.js
@@ -88,14 +88,18 @@ if (opts['server-option']) {
 
 (async () => {
   if (opts._.length) {
-    opts.spec = opts._;
+    if (opts._[0] === '-') {
+      opts.spec = process.stdin;
+    } else {
+      opts.spec = opts._;
+    }
   }
   delete opts._;
   try {
     const { exit_code } = await mochify(opts);
-    process.exitCode = exit_code;
+    process.exitCode = exit_code; // eslint-disable-line require-atomic-updates
   } catch (e) {
     console.error(e.stack);
-    process.exitCode = 1;
+    process.exitCode = 1; // eslint-disable-line require-atomic-updates
   }
 })();
--- a/cli/index.js
+++ b/cli/index.js
@@ -86,20 +86,19 @@ if (opts['server-option']) {
   opts.server_options = opts['server-option'];
 }
 
-(async () => {
-  if (opts._.length) {
-    if (opts._[0] === '-') {
-      opts.spec = process.stdin;
-    } else {
-      opts.spec = opts._;
-    }
+if (opts._.length) {
+  if (opts._[0] === '-') {
+    opts.spec = process.stdin;
+  } else {
+    opts.spec = opts._;
   }
-  delete opts._;
-  try {
-    const { exit_code } = await mochify(opts);
-    process.exitCode = exit_code; // eslint-disable-line require-atomic-updates
-  } catch (e) {
-    console.error(e.stack);
-    process.exitCode = 1; // eslint-disable-line require-atomic-updates
-  }
-})();
+}
+delete opts._;
+mochify(opts)
+  .catch((err) => {
+    console.error(err.stack);
+    return { exit_code: 1 };
+  })
+  .then(({ exit_code }) => {
+    process.exitCode = exit_code;
+  });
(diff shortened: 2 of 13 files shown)


cli/index.js
```diff
@@ -94,14 +94,18 @@ if (opts['server-option']) {
...
+      opts.spec = process.stdin;
+    } else {
+      opts.spec = opts._;
+    }
   }
   delete opts._;
   try {
     const { exit_code } = await mochify(opts);
-    process.exitCode = exit_code;
+    process.exitCode = exit_code; // eslint-disable-line require-atomic-updates
```
m90, line 106: I have no idea why, but `eslint` now thinks this is a possible race condition??

cli/test/puppeteer.integration.js
```diff
@@ -28,6 +29,27 @@ describe('puppeteer', () => {
     assert.equals(json.tests[0].fullTitle, 'test passes');
   });
 
+  it('reads from stdin', async () => {
+    let result;
+    try {
+      const cp = execa('../../index.js', ['--driver', 'puppeteer', '-'], {
```
m90, line 35: Alternatively we could also teach `run` about this behaviot, dunno if it's worth it?

mochify/lib/load-config.js
```diff
@@ -10,19 +10,42 @@ async function loadConfig(options) {
...
     const specified = await explorer.load(options.config);
     const config = Object.assign(specified.config, options);
-    return mergeWithDefault(default_config_promise, config);
+    merged = await mergeWithDefault(default_config_promise, config);
+  } else {
+    merged = await mergeWithDefault(default_config_promise, options);
   }
 
-  return mergeWithDefault(default_config_promise, options);
+  const validation_error = validate(merged);
```
m90, line 22: Is this the right place to introduce something like this?

mochify/lib/load-config.js
```diff
@@ -10,19 +10,42 @@ async function loadConfig(options) {
...
+  }
+  return merged;
 }
 
 async function mergeWithDefault(default_config_promise, config) {
   const default_config = await default_config_promise;
   if (default_config) {
-    return deepmerge(default_config.config, config);
+    return deepmerge(default_config.config, config, {
+      clone: false
```
m90, line 33: Cloning a stream apparently causes infinite recursion. Don't think not cloning will hurt us here.

mochify/lib/resolve-spec.js
```diff
@@ -6,6 +6,10 @@ const glob = promisify(require('glob'));
 exports.resolveSpec = resolveSpec;
 
 async function resolveSpec(spec = 'test/**/*.js') {
+  if (typeof spec === 'object' && typeof spec.pipe === 'function') {
```
m90, line 9: I found it nicer to handle such situations at `lib` level instead of `index.js` as index.js already contains a lot of control flow and it would have gotten even more.

m90:
I seem to have broken this with some last minute tweaks, will make it a draft again until it's fixed.

mochify/lib/load-config.js
mantoni: I'd put the validation function into its own file and move the validate call outside of `loadConfig`.

mochify/lib/resolve-spec.js
mantoni: Yeah, I also get the feeling it's growing quite a bit.

Maybe it's time to introduce more functions like `run` that handle parts of the control flow. E.g. `configure`, `launch`, `run` and `shutdown`. Just a random morning thought though 🙃

cli/index.js
m90: I filed an issue here https://github.com/eslint/eslint/issues/15076

mochify/lib/validate-config.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+exports.validateConfig = validateConfig;
+
+function validateConfig(config) {
```
m90, line 5: I left this function intentionally simple (i.e. manual checking) for now. We can start building our declarative mini DSL later ...

cli/index.js
```diff
@@ -86,16 +86,19 @@ if (opts['server-option']) {
...
+    opts.spec = process.stdin;
+  } else {
     opts.spec = opts._;
   }
-  delete opts._;
-  try {
-    const { exit_code } = await mochify(opts);
+}
+delete opts._;
+mochify(opts)
```
m90, line 97: Unwrapping this and using a Promise chain is here to stop eslint from complaining. I'm not entirely sure if its complaints are correct or not yet. We could also keep the old version and ignore the rule, no real preference from my side here.

See https://github.com/eslint/eslint/issues/15076

cli/README.md
```diff
@@ -64,3 +64,21 @@ Files in the given directory will be served as static assets.
...
 Currently only `--server-option.port` for passing the port to use is supported.
+
+## Spec
+
+The `spec` argument can be a list of files or a glob pattern that will be resolved by Mochify.
+
+```
+mochify ./src/foo.test.js ./src/bar.test.js
+mochify ./src/*.test.js # Let the shell handle glob expansion
+mochify "./src/*.test.js" # Let Mochify handle glob expansion
```
mantoni, line 75: 👍

cli/index.js
mantoni: Probably even easier to read than the self-invoking async function. 👍

mochify/index.js
```diff
@@ -14,11 +15,16 @@ exports.mochify = mochify;
 async function mochify(options = {}) {
   const config = await loadConfig(options);
 
+  const validation_error = validateConfig(config);
+  if (validation_error) {
+    throw validation_error;
+  }
```
mantoni, line 21: Why is `validateConfig` not throwing the errors directly?

mantoni (approved):
Looks very good to me. Thank you. Just a minor stylistic question, but feel free to merge.

cli/test/puppeteer.integration.js
mantoni: That'd make it rather complicated, no?

mochify/index.js
m90: Looks like I've just been writing too much Go lately. I'll change this as it makes sense for this to behave like everything else (which is throwing to stop execution).

PR #270, opened 2021-09-16
Unnest yargs definition to restore argv._

m90:
It seems #267 introduced a subtle breakage: it seems yargs now somehow parses `spec` by itself, rendering the logic around `opts._` unneccessary. This would be fine all by itself (we could remove https://github.com/mantoni/mochify.js/blob/b407d82bda319251e7dcda039bbbd4a273d6224b/cli/index.js#L96-L98) until #268 enters the stage: when handling positional arguments yargs will just eat `-` without any further notice: https://github.com/yargs/yargs/issues/1312

Why did I introduce the nested command in #267 the first place? It seems that if you pass a second argument to `.usage()` yargs will somehow not display `example`, `epilogue` and such unless you start attaching it to the child instance, which means this PR needs to drop the command description again. I don't know if this is a bug in yargs, I am doing something wrong or something completely different is going on.

This is also the reason the tests for #269 are currently broken (after I rebased against `rewrite` that is).

Sorry if this is terribly confusing.

### DIFF
--- a/cli/index.js
+++ b/cli/index.js
@@ -6,33 +6,7 @@ const { hideBin } = require('yargs/helpers');
 const { mochify } = require('@mochify/mochify');
 
 const opts = yargs(hideBin(process.argv))
-  .usage(
-    '$0 [options] <spec...>',
-    'Run Mocha tests in real browsers.',
-    (cmd) => {
-      cmd
-        .example(
-          '$0 --driver puppeteer --bundle browserify "./src/**/*.test.js" ',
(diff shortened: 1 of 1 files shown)


mantoni (approved):
I don't really understand, but I trust you do 😉

m90:
yargs API is mega confusing, I don't understand too much either.

I think the basic problem can be described as: if we want to allow passing `-` for positionals we need to prevent yargs from handling these at all (which is why they still end up in `_`) because it just doesn't support `-` as a value. And there seem to many, highly surprising ways in which you can make yargs start handling positionals (e.g. provide usage info) ...

PR #271, opened 2021-09-20
Allow passing launch options to Playwright, add docs

m90:
Basically the same as #252 but for `driver-playwright`

Also tweaks the puppeteer setup a little.


mantoni (approved):
👍

PR #272, opened 2021-09-22
Allow passing down options to JSDOM

m90:
#252, but for `driver-jsdom`

### DIFF
--- a/driver-jsdom/index.js
+++ b/driver-jsdom/index.js
@@ -5,17 +5,24 @@ const jsdom = require('jsdom');
 exports.mochifyDriver = mochifyDriver;
 
 function mochifyDriver(options = {}) {
-  const stderr = options.stderr || process.stderr;
+  const {
+    stderr = process.stderr,
+    url = 'http://localhost',
+    pretendToBeVisual = true,
+    strictSSL = false,
+    ...jsdom_options
+  } = options;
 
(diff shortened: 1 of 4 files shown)


driver-jsdom/index.js
```diff
@@ -5,17 +5,24 @@ const jsdom = require('jsdom');
...
-      url: options.url || 'http://localhost',
+      url,
+      pretendToBeVisual,
+      strictSSL,
+      ...jsdom_options,
       virtualConsole: virtual_console,
-      runScripts: 'dangerously',
-      pretendToBeVisual: true,
-      strictSSL: false
+      runScripts: 'dangerously'
```
m90, line 25: Placing these two mandatory settings down here makes sure that it works, but might trip up people that don't read the documentation. Should this rather be allowed to pass through as well, even if it makes Mochify fail?

mantoni: If we don't put it here, wouldn't everyone have to provide this as an option for the driver to work? All drivers should work out of the box – at least to some extend.

mantoni (approved):

driver-jsdom/index.js
m90: No, what I meant is allowing consumers to override these in the first place. Right now, you could pass your own runScripts value, but Mochify would still use its preset.

PR #273, opened 2021-09-22
Allow driver option to be a module on API level

m90:
Consumers of the API module might want to pass in a custom (?) driver module instead of having to rely on Mochify to do the resolution:

```js
mochify({
  driver: require('./my-driver-thing'),
  ...other
})
```

---

While this might look simple in the above snippet, there is one thing I am not sure about: does this mean that you're using a `.js` config file, you could require the driver in there and pass it down? In case yes, do we want that or should config stay rather dumb?

### DIFF
--- a/mochify/index.js
+++ b/mochify/index.js
@@ -73,22 +73,27 @@ async function mochify(options = {}) {
   return { exit_code };
 }
 
-function resolveMochifyDriver(name) {
-  let driverModule;
-  try {
-    // eslint-disable-next-line node/global-require
-    driverModule = require(`@mochify/driver-${name}`);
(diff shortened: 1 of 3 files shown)


mantoni:
Thank you for doing all the work on the rewrite lately 🙏.

Not sure about the .js config idea. Let's worry about that once someone runs into it.

There's quite a bit of logic in the module resolution code now. Can you add some unit tests?

m90:
> There's quite a bit of logic in the module resolution code now. Can you add some unit tests?

Makes sense, I moved the function into its own module and added unit tests.

mantoni (approved):
🙌

PR #274, opened 2021-09-23
Run integration tests for driver-webdriver against selenium server

### LINKED ISSUE #246: Rewrite: Integration tests for WebDriver
Integration tests for WebDriver could launch a local Selenium server and run some tests against it, like the `cli` integration tests.

We could also think about running tests against SauceLabs.

m90:
Closes #246 

I rarely run Selenium locally, so I am not sure what the most common setups are here. I tested this using the same Docker image as in the CI setup and it worked fine.

Is it ok to skip the tests if not in CI and there is no server running?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,5 +42,17 @@ jobs:
         run: npm run prettier:check
       - name: Test
         run: npm test
+      # N.B.: The selenium server cannot run as a GitHub Actions "service" as
+      # the tests need to mount local data into the container.
+      - name: Start Selenium Server
+        run: |
+          docker run -d -p 4444:4444 \
+            --shm-size="2g" \
+            --name selenium_server \
+            -v ${PWD}:${PWD} \
+            selenium/standalone-firefox:4.0.0-rc-2-prerelease-20210916
       - name: Integration
         run: npm run test:integration
+      - name: Stop Selenium Server
+        if: always()
+        run: docker stop selenium_server && docker rm selenium_server
--- a/cli/test/webdriver.integration.js
+++ b/cli/test/webdriver.integration.js
@@ -21,8 +21,14 @@ describe('webdriver', () => {
     }
   }
 
+  let skip;
+  before(async () => {
+    const selenium_running = await pingSelenium();
+    skip = !process.env.CI && !selenium_running;
+  });
+
   it('passes', async function () {
-    if (!process.env.CI && !(await pingSelenium())) {
+    if (skip) {
       this.skip();
     }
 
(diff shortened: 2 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -42,5 +42,17 @@ jobs:
         run: npm run prettier:check
       - name: Test
         run: npm test
+      # N.B.: The selenium server cannot run as a GitHub Actions "service" as
+      # the tests need to mount local data into the container.
+      - name: Start Selenium Server
```
m90, line 47: This adds roughly 30s to every run, which seems ok, but also not super fast. Do you know of a faster way to get this up and running?

.github/workflows/test.yml
```diff
@@ -42,5 +42,17 @@ jobs:
...
+        run: |
+          docker run -d -p 4444:4444 \
+            --shm-size="2g" \
+            --name selenium_server \
+            -v ${PWD}:${PWD} \
```
m90, line 52: So I am not sure if this is expected or not, but the driver only works if it has access to the local FS in order to read the `index.html` file. Is this ok for SauceLabs / BrowserStack et. al. scenarios?

.github/workflows/test.yml
```diff
@@ -42,5 +42,17 @@ jobs:
...
+            selenium/standalone-firefox:4.0.0-rc-2-prerelease-20210916
       - name: Integration
         run: npm run test:integration
+      - name: Stop Selenium Server
+        if: always()
```
m90, line 57: I created a forceful failure and this seems to run as expected: https://github.com/mantoni/mochify.js/actions/runs/1265187816

.github/workflows/test.yml
m90: It seems GitHub Actions keeps some Docker images in a cache (which is nice), but they won't tell us which ones (or at least I can't find it): https://github.com/actions/virtual-environments/blob/main/docs/software-and-images-guidelines.md#software-and-images-support-policy

cli/test/webdriver.integration.js
```diff
@@ -0,0 +1,116 @@
...
+  let skip;
+  before(async () => {
+    const selenium_running = await pingSelenium();
+    skip = !process.env.CI && !selenium_running;
+  });
+
+  it('passes', async function () {
+    if (skip) {
+      this.skip();
+    }
```
mantoni, line 33: I think this can be done in `before` to skip all tests.

.github/workflows/test.yml
mantoni: I was thinking, if we have a website hosted on GitHub pages, we could have the `index.html` file there. It could even be the default URL in case the host is not `localhost`.

.github/workflows/test.yml
mantoni: We can always optimize for performance later. It's great to have this in place 👍

mantoni (approved):
Just one minor suggestion. Thank you for setting this up!

PR #275, opened 2021-09-24
Wrap webdriver script content in IIFE, removing need for evaluateReturn

m90:
Picks up: https://github.com/mantoni/mochify.js/pull/232/files/5aecae9f8be4f783f27877808fe4496481cc71dd#r667336402

Just prepending `return` seems to limit ourselves to immediately return the first expression in the given script, but wrapping the script in an IIFE seems to work as expected.

This should probably merged after #274 so it can be rebased against `rewrite` before and pass the tests (it does pass locally for me).

### DIFF
--- a/driver-webdriver/index.js
+++ b/driver-webdriver/index.js
@@ -29,8 +29,11 @@ async function mochifyDriver(options = {}) {
 
   return {
     evaluate: (script) =>
-      client.executeScript(
-        `return (function () { var __mochify_return_value = ${script}; return __mochify_return_value; })()`,
+      client.executeAsyncScript(
+        `
+          var __mochify_return_value = ${script};
+          arguments[arguments.length - 1](__mochify_return_value)
+        `,
         []
       ),
     end: () => client.deleteSession()
--- a/driver-webdriver/index.js
+++ b/driver-webdriver/index.js
@@ -28,8 +28,11 @@ async function mochifyDriver(options = {}) {
   await client.navigateTo(url || default_url);
 
   return {
-    evaluate: (script) => client.executeScript(script, []),
-    evaluateReturn: (script) => client.executeScript(`return ${script}`, []),
+    evaluate: (script) =>
+      client.executeScript(
+        `return (function () { var __mochify_return_value = ${script}; return __mochify_return_value; })()`,
+        []
+      ),
     end: () => client.deleteSession()
   };
 }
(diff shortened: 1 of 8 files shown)


driver-webdriver/index.js
```diff
@@ -28,8 +28,11 @@ async function mochifyDriver(options = {}) {
   await client.navigateTo(url || default_url);
 
   return {
-    evaluate: (script) => client.executeScript(script, []),
-    evaluateReturn: (script) => client.executeScript(`return ${script}`, []),
+    evaluate: (script) =>
+      client.executeScript(
+        `return (function () { var __mochify_return_value = ${script}; return __mochify_return_value; })()`,
```
m90, line 33: I'm not entirely sure if the wrapping IIFE is even needed. This also works if you do

```js
var __mochify_return_value = ${script}; return __mochify_return_value;
```

yet I have a hard time understanding if each `executeScript` call creates a function scope by itself or not and if this could create possible collisions.

See: https://w3c.github.io/webdriver/#dfn-execute-script

m90: Thinking about it, if it's ok to use a `return` statement, this must be run in the context of a function, so it's probably fine to omit the wrapper.

mantoni: AFAIK some implementations warp the code in a function and some don't.

We also use this to inject the mochify client, which I guess works because that is wrapped in a function too.

I'll run some tests as well this weekend.

mantoni:
From my investigation you're right, all injected scripts are wrapped in a function call. Therefore the wrapper isn't needed and we can simply use `return` directly.

I've allowed myself to add two commits to this PR.

1) The first one is a refinement / alternative to the `return` based implementation, where I'm now calling `executeAsyncScript`. The "invisible" function wrapped used in these invocations received a callback as the last argument. I'm also happy to go with a simple synchronous `return` instead, just wanted to let you know. What do you think?
2) The second moves wrapping the client in a self-invoking function into `setup-client`, so mostly whitespace changes. I figured this makes the client easier to read and we can control the final semicolon.

m90:
Looks good to me as well, that makes it pretty clear what's going on now. I'd have no objections against merging this.

m90:
I rebased the branch so it also contains the integration tests for webdriver, seems everything is passing.

mantoni:
Great! Thank you 🙏

PR #276, opened 2021-12-23
Update mocaccino to v6 and mocha to v9

albertyw:
Pulls in mocaccino v6 and mocha v9.  Mocha v9 is used for internal testing of mochify.

Depends on https://github.com/mantoni/mocaccino.js/pull/38 to update mocaccino versions.


package.json
```diff
@@ -85,7 +85,7 @@
   "devDependencies": {
     "@studio/changes": "^2.0.1",
     "eslint": "^7.13.0",
-    "mocha": "^8.4.0",
+    "mocha": "^9.1.0",
```
m90, line 88: Not sure if this your intention in this PR, but this is the Mocha version used by the tests in this repo. Mochify itself will still be on version 8 as per the version of mocaccino that is used. Is that the idea here?

albertyw: I wanted to check if mochify.js would be compatible with mocha v9, as per https://github.com/mantoni/mocaccino.js/pull/38#discussion_r774395444.

albertyw: I added another commit to also update mocaccino (temporarily to my fork) to see if there would be any conflicts with mocha.

m90 (approved):
LGTM. Are you ok if we make this `9.0.0` @mantoni ?

mantoni (approved):
@m90 Go! ❤️

m90:
This is released as `9.0.0` :nail_care:

PR #279, opened 2022-02-20
Update puppeteer to v13

albertyw:
This updates to the latest version of puppeteer to get rid of annoying security vulnerability warnings.  There are multiple major version changes but none of the breaking changes seem to affect mochify.  

See [puppeteer's changelogs](https://github.com/puppeteer/puppeteer/blob/main/CHANGELOG.md).


m90:
Thanks for providing the link to the changelog. I also wouldn't be worried about any of those changes, so I released this as `9.2.0`. Thanks again :100:

